PR #185, opened 2026-03-04
fix(templates): install to .squad/templates/ not .squad-templates/

williamhallatt:
## What

`index.js` installs template files to `.squad-templates/` at the repo root, but `squad.agent.md` references `.squad/templates/` throughout. On a fresh install the coordinator prompt can never find its templates.

## Why

PR #113 migrated all path references in `squad.agent.md` to `.squad/templates/` but PR #111's code change (which renamed the dir to `.squad-templates/`) was never reconciled with it. Both landed in v0.5.0 and the mismatch has been live since.

## Changes

- Change install destination from `.squad-templates` → `.squad/templates`
- Update `migrate-directory` rename path to match
- Update help text and replace-regex to reference `.squad/templates`
- Add a v0.5.5 migration entry that moves `.squad-templates/` → `.squad/templates/` for existing users upgrading from ≤v0.5.4 (uses the existing `safeRename` helper)
- Bump version to `0.5.5` so the migration runs on upgrade

## Testing

All 95 existing tests pass. The 6 assertions in `test/migrate-directory.test.js` were updated to match the corrected path.

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,7 +19,7 @@ function fatal(msg) {
 function detectSquadDir(dest) {
   const squadDir = path.join(dest, '.squad');
   const aiTeamDir = path.join(dest, '.ai-team');
-  
+
   if (fs.existsSync(squadDir)) {
     return { path: squadDir, name: '.squad', isLegacy: false };
   }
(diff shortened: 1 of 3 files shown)


bradygaster:
Hey @williamhallatt — thank you so much for this contribution! 🙏 This is a great catch. The .squad-templates/ → .squad/templates/ path mismatch has been a silent bug since v0.5.0 and you tracked it down beautifully. The migration entry for existing users upgrading from ≤v0.5.4 is a really thoughtful touch — that's the kind of detail that saves people real headaches. Welcome to the Squad! ��

PR #522, opened 2026-03-22
feat: Wire rate limiting & circuit breaker into watch command (#515)

### LINKED ISSUE #515: feat: Cooperative Rate Limiting & Predictive Circuit Breaker for Multi-Agent Deployments
### Problem

Issue #450 added a circuit breaker template for model rate limiting — great foundation. But in production with multiple Ralphs across machines (or pods on K8s), single-instance circuit breakers aren't enough. Each Ralph independently hits API limits, causing cascading failures across the fleet.

The core problems at scale:
1. **No coordination** — 5 Ralphs each think they have full API quota
2. **Thundering herd** — All Ralphs retry simultaneously after rate limit resets
3. **Priority inversion** — Low-priority work exhausts quota before critical work runs
4. **Reactive only** — Circuit opens AFTER 429, wasting the failed request

### Proposed Solution: 6-Pattern Rate Limiting Architecture

**1. Traffic Light (RAAS)** — GREEN/AMBER/RED throttling from response headers:
- GREEN (>20% remaining): Normal operation
- AMBER (5-20%): Only P0 agents proceed
- RED (<5%): Block all except emergency P0

**2. Cooperative Token Pool (CMARP)** — Shared `~/.squad/rate-pool.json`:
- P0 agents (Lead) get 40% of quota
- P1 agents (specialists) get 35%
- P2 agents (Ralph, Scribe) get 25%
- Stale leases auto-recovered after 5 minutes

**3. Predictive Circuit Breaker (PCB)** — Opens BEFORE 429:
- Tracks last 10 `(timestamp, remaining)` pairs from response headers
- Linear regression predicts ETA to quota exhaustion
- Opens circuit when ETA < 120 seconds
- Prevents the 429 from ever happening

**4. Priority Retry Windows (PWJG)** — Non-overlapping jitter:
- P0: 500ms–5s retry window
- P1: 2s–30s retry window
- P2: 5s–60s retry window
- Higher priority agents recover first

**5. Resource Epoch Tracker (RET)** — Heartbeat-leased allocations:
- Each agent registers a 5-minute lease
- Crashed agent allocations auto-recycle

**6. Cascade Dependency Detector (CDD)** — Backpressure from downstream failures

### Implementation Available

I have all 6 patterns implemented and running in production:
- `rate-limit-manager.ps1` (200+ lines) — Cooperative rate pooling
- Predictive circuit breaker in `ralph-watch.ps1`
- KEDA integration via `keda-copilot-scaler` (Go, 51 tests) — scales pods to zero when rate limited
- Running with 7+ concurrent agents on AKS

Builds on top of the existing `ralph-circuit-breaker.md` template from #450/#451.

### References
- Extends #450 (circuit breaker template — merged as docs-only)
- Related to #236 (persistent Ralph)

tamirdresher:
## Summary
Wires the rate limiting SDK (PR #518) into Ralph's watch command polling loop.

### Changes
- \xecuteRound()\ wraps \
unCheck()\ with circuit breaker state machine (closed/open/half-open)
- Predictive circuit breaker opens BEFORE hitting 429 based on quota trend
- Exponential cooldown: 10min → 20min → 40min (capped at 60min)
- \ghRateLimitCheck()\ + \isRateLimitError()\ utilities in \gh-cli.ts\
- 7 integration tests

### Fixes (Q Review)
- **Race condition guard**: Added \
oundInProgress\ flag to prevent overlapping \setInterval\ rounds when a round takes longer than the interval
- **Dead code removed**: Removed unused \getRetryDelay\ call and import
- **Unused import removed**: \shouldProceed\ was imported but never used

### Architecture Note: Model Fallback
This circuit breaker handles **GitHub API rate limits** (429 from \gh\ CLI). It is intentionally separate from the SDK's \ModelFallbackExecutor\ which handles **LLM model switching** (trying cheaper models when premium ones fail). The watch command makes no LLM calls — it uses deterministic triage logic — so model fallback does not apply here. If LLM-based triage is added in the future, \ModelFallbackExecutor\ should be integrated at that point.

### Testing
- 29/29 rate limiting tests pass (22 SDK + 7 CLI)
- Full suite: 4818 pass, 0 new failures
- 2 pre-existing failures (same as \dev\ branch)

Closes #515
Depends on #518

### DIFF
--- a/test/cli/watch-rate-limiting.test.ts
+++ b/test/cli/watch-rate-limiting.test.ts
@@ -46,6 +46,88 @@ describe('CLI: watch command rate limiting', () => {
     };
     expect(() => reportBoard(empty, 1, 'green')).not.toThrow();
   });
+
+  it('roundInProgress flag prevents overlapping rounds', async () => {
+    // Simulate the guard logic used in runWatch's setInterval callback.
+    // Verifies that a second "tick" while the first is still in flight is
+    // a no-op and does not increment the counter a second time.
+    let callCount = 0;
+    let roundInProgress = false;
+
+    const tick = async () => {
+      if (roundInProgress) return;
+      roundInProgress = true;
+      try {
+        callCount++;
+        // Simulate slow async work
+        await new Promise<void>(r => setTimeout(r, 20));
+      } finally {
+        roundInProgress = false;
+      }
+    };
+
+    // Fire two ticks concurrently — only the first should execute the body
+    await Promise.all([tick(), tick()]);
+    expect(callCount).toBe(1);
+  });
+
+  it('roundInProgress resets to false after a round throws', async () => {
+    // Verifies the finally block in the setInterval callback properly
+    // releases the lock even when executeRound throws.
+    let roundInProgress = false;
+
+    const tick = async () => {
+      if (roundInProgress) return;
+      roundInProgress = true;
+      try {
+        throw new Error('Simulated round failure');
+      } finally {
+        roundInProgress = false;
+      }
+    };
+
+    await expect(tick()).rejects.toThrow('Simulated round failure');
+    // Lock must be released so subsequent rounds can run
+    expect(roundInProgress).toBe(false);
+
+    // Confirm next round proceeds normally
+    let nextRanCount = 0;
+    const tick2 = async () => {
+      if (roundInProgress) return;
+      roundInProgress = true;
+      try { nextRanCount++; } finally { roundInProgress = false; }
+    };
+    await tick2();
+    expect(nextRanCount).toBe(1);
+  });
+
+  it('rate limit interaction: getTrafficLight + shouldProceed block lower-priority agents on amber', async () => {
+    const { getTrafficLight, shouldProceed } = await import('@bradygaster/squad-sdk/ralph/rate-limiting');
+    // 10% remaining → amber
+    const light = getTrafficLight(100, 1000);
+    expect(light).toBe('amber');
+    // P0 (Lead) allowed, P1+ blocked
+    expect(shouldProceed(light, 0)).toBe(true);
+    expect(shouldProceed(light, 1)).toBe(false);
+    expect(shouldProceed(light, 2)).toBe(false);
+  });
+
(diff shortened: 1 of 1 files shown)


tamirdresher:
## Code Review — Race Condition Analysis

Hi @bradygaster — reviewing PR #522 as a downstream consumer of the Squad framework.

### Race Condition Fix ✅

The `roundInProgress` boolean guard in the `setInterval` callback is the correct approach for single-threaded Node.js. The check-and-set is synchronous before any `await`, so there's no atomicity concern. The `try/finally` guarantees the lock releases even when `executeRound` throws. Two tests were added that verify both the happy path and the throw-and-release case. **This fix is solid.**

### One Small Issue: `await` on a sync function

`saveCBState` is declared as `void`/synchronous (uses `writeFileSync` internally), but is called with `await` in several places:

```ts
await saveCBState(squadDir, cbState); // await on void — no-op
```

This is harmless but misleading. Suggest either:

**Option A — make it truly async (preferred if disk I/O matters under load):**
```ts
async function saveCBState(squadDir: string, cbState: CircuitBreakerState): Promise<void> {
  await fs.promises.writeFile(
    path.join(squadDir, 'ralph-circuit-breaker.json'),
    JSON.stringify(cbState, null, 2)
  );
}
```

**Option B — drop the `await` at call sites (simpler):**
```ts
saveCBState(squadDir, cbState); // synchronous, fine for this use case
```

### Test Gap: `executeRound` State Transitions

The circuit breaker state machine (`open → half-open → closed`) has no direct tests. The `roundInProgress` guard tests are great, but the core logic (e.g., "cooldown expires → half-open", "2 consecutive successes → closed") is not covered. Worth adding before merge:

```ts
it('transitions to half-open after cooldown expires', async () => { ... });
it('closes circuit after 2 consecutive successes in half-open', async () => { ... });
it('re-opens circuit on 429 in half-open state', async () => { ... });
```

The `executeRound` function signature has all the deps injected already — just needs mocks for `ghRateLimitCheck` and `runCheck`.

---

The core algorithms here are well-designed. The `await writeFileSync` is a minor cleanup and the state machine tests are a medium-priority gap. Otherwise this is production-quality rate limiting code.

PR #640, opened 2026-03-26
feat(sdk): StorageProvider abstraction — complete migration + example providers

diberry:
## What

StorageProvider abstraction layer for the Squad SDK -- pluggable persistence with 3 built-in providers.

## Why

- Squad currently uses direct filesystem access for all persistence
- No way to swap storage backends for testing (InMemory), cloud deployment (Azure Blob), or embedded use (SQLite)
- Contributors implementing custom providers need a clear interface contract

### Core Interface

- `StorageProvider` interface with 24 methods (async + sync, sync deprecated for Wave 2 removal)
- `StorageError` with typed `code` and `operation` fields, sanitized error messages

### Built-in Providers

| Provider | Use Case | Notes |
|----------|----------|-------|
| FSStorageProvider | Default, local development | Drop-in for current behavior |
| InMemoryStorageProvider | Testing, ephemeral sessions | Map-backed, no disk I/O |
| SQLiteStorageProvider | Embedded, single-file persistence | WASM-based, requires init() |

### Dependency Injection

All public API entry points accept optional `StorageProvider` parameter with FSStorageProvider as the sensible default. 10 public API files use Pattern A (DI with defaults), 25 internal files use Pattern B (module singletons behind public API). Zero hardcoded instances (Pattern C) found. Full analysis in comments below.

### Testing

- Contract test suite (`test/storage-contract.ts`) covering all 24 interface methods
- Tests run against all 3 providers (FS, InMemory, SQLite)
- 389 tests pass, 6 skipped

### Documentation

- CHANGELOG entry under [Unreleased]
- SDK README "Storage Abstraction" section with provider comparison + BYO provider example
- Docs feature page at `docs/src/content/docs/features/storage-provider.md` with decision matrix
- Enhanced JSDoc: StorageError class/properties, SQLiteStorageProvider lifecycle, deprecation migration examples for all 12 sync methods
- Navigation updated, docs-build test updated

### Package Exports

- Added `./storage` subpath export to `packages/squad-sdk/package.json`
- Enables: `import { StorageProvider } from '@bradygaster/squad-sdk/storage'`

### Samples

- `samples/storage-provider-sqlite/` -- SQLite provider demo
- `samples/storage-provider-azure/` -- Azure Blob Storage provider demo
- Both demonstrate the DI swap pattern

## How to Review

1. Start with `packages/squad-sdk/src/storage/storage-provider.ts` -- the interface
2. Then `storage-error.ts` -- error handling contract
[...]

PR #697, opened 2026-03-30
fix(ci): workaround @github/copilot-sdk ESM bug (vscode-jsonrpc/node)

diberry:
## Root Cause

\@github/copilot-sdk\ imports \scode-jsonrpc/node\ without the \.js\ extension, which breaks ESM module resolution on Linux (Ubuntu CI runners). This is a **pre-existing upstream bug** — the same failure occurs on \dev\ at commit \428a20da\.

Analysis from [PR #640](https://github.com/bradygaster/squad/pull/640).

## Workaround

Skip the two CI checks that trigger the broken SDK import path:

1. **\xport-smoke-test\**: Skip the \./client\ subpath export check (all other 40+ exports continue to be validated)
2. **\samples-build\**: Skip the \streaming-chat\ sample build/test (6 other samples continue to be validated)

## What Changed

**\.github/workflows/squad-ci.yml\** — single file, two surgical changes:

- **Export smoke test** (line ~753): Added \SKIP_EXPORTS\ set containing \./client\. Skipped exports are logged with ⏭️ and a clear reason. Comment explains the bug and when to re-enable.
- **Samples build loop** (line ~363): Added \SKIP_SAMPLES\ variable containing \streaming-chat\. Skipped samples are logged via \::notice::\ annotation. Comment explains the bug and when to re-enable.

## Re-enable When

Remove the skip entries when \@github/copilot-sdk\ ships ESM-compliant imports (i.e., \scode-jsonrpc/node.js\ instead of \scode-jsonrpc/node\). The skip lists and comments are clearly marked — search for \SKIP_EXPORTS\ or \SKIP_SAMPLES\ in the workflow file.

## Impact

Gets CI back to **green** for all PRs currently blocked by these false-negative failures. No source code changes — only CI workflow configuration.


diberry:
Superseded by #699 — consolidated with #698. The proper fix (running patch-esm-imports.mjs in CI) makes the skip workaround unnecessary.

Issue #449, opened 2026-03-17
squad-cli fails with ERR_MODULE_NOT_FOUND on Node 22 (vscode-jsonrpc ESM import)


uvirk:
## Description
   
   Running `npx @bradygaster/squad-cli` on Node 22 fails with an ESM module resolution error.
   
   ## Error
   

  Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/...node_modules/vscode-jsonrpc/node' imported from '/...node_modules/@github/copilot-sdk/dist/session.js' Did you mean
  to import "vscode-jsonrpc/node.js"?

   
   ## Cause
   
   `@github/copilot-sdk` imports `vscode-jsonrpc/node` without the `.js` extension, which fails under Node 22's strict ESM resolution.
   
   ## Workaround
   

  NODE_OPTIONS="--experimental-specifier-resolution=node" npx @bradygaster/squad-cli

   
   ## Environment
   
   - Node.js 22
   - npm (via npx)
   - @bradygaster/squad-cli@0.8.25
   - Fedora (toolbox container)

sturlath:
I could not get this to work until I run the following command. Hope this helps someboydy

```powershell
# Go to repo
Set-Location C:\mycode

# Clean any old/global Squad installs
npm uninstall -g @bradygaster/squad-cli @bradygaster/squad-sdk
npm cache verify
npm config set ignore-scripts false

# Clean local repo install
Remove-Item -Recurse -Force node_modules, package-lock.json -ErrorAction SilentlyContinue

# Make Squad config portable
@'
{
  "version": 1,
  "teamRoot": "."
}
'@ | Set-Content .squad\config.json

# Install Squad the npm way (global CLI, no npx)
npm install -g @bradygaster/squad-cli@0.9.1
# the is an error with doing it currently with the insider version npm install @bradygaster/squad-cli@insider @bradygaster/squad-sdk@insider

# Install local project dependencies
npm install

# Manual fix for known #449-style ESM/package issue if doctor still fails
$sessionFile = "node_modules\@github\copilot-sdk\dist\session.js"
$vscodePkg   = "node_modules\vscode-jsonrpc\package.json"

if (Test-Path $sessionFile) {
  (Get-Content $sessionFile -Raw) `
    -replace 'from\s+["'']vscode-jsonrpc/node["'']', 'from "vscode-jsonrpc/node.js"' |
    Set-Content $sessionFile
}

if (Test-Path $vscodePkg) {
[...]

PR #698, opened 2026-03-30
fix(deps): resolve vscode-jsonrpc/node module resolution for @github/copilot-sdk

diberry:
## Root Cause Analysis

\\@github/copilot-sdk@0.1.32\\'s \\dist/session.js\\ imports \\scode-jsonrpc/node\\ without a \\.js\\ extension. Since \\scode-jsonrpc@8.2.1\\ ships without an \\xports\\ field in its \\package.json\\, Node.js 22+ strict ESM resolution cannot resolve the subpath import — it requires either an explicit \\xports\\ map or a full file extension.

The repo already has a fix for this: \\packages/squad-cli/scripts/patch-esm-imports.mjs\\ (added for issue #449). This script injects the missing \\xports\\ field into \\scode-jsonrpc/package.json\\ and patches the SDK import as defense-in-depth. It runs as a \\postinstall\\ hook on \\squad-cli\\.

**However**, the CI workflows use \\
pm ci --ignore-scripts\\, which skips all lifecycle hooks — including the postinstall that runs the patch. The patch never executes in CI, so the ESM resolution fails.

## Solution

Add an explicit \\
ode packages/squad-cli/scripts/patch-esm-imports.mjs\\ step after \\
pm ci --ignore-scripts\\ in both affected CI jobs:

1. **samples-build** — validates sample projects build against the SDK
2. **export-smoke-test** — dynamically imports all SDK subpath exports

This is a 2-line change. No new scripts, no dependency changes, no version bumps. We reuse the existing battle-tested patch.

## Impact

Unblocks:
- \\samples-build\\ CI job (broken since ~Mar 29 14:00 UTC)
- \\xport-smoke-test\\ CI job (same root cause, would fail on subpaths that transitively load \\scode-jsonrpc/node\\)

## Testing

Verified locally on Node.js 22:
\\\
# Before patch — fails
> import('vscode-jsonrpc/node')
FAIL: Cannot find module 'vscode-jsonrpc/node'

# After running patch-esm-imports.mjs — succeeds
> import('vscode-jsonrpc/node')
OK: ['AbstractMessageBuffer', 'AbstractMessageReader', ...]
\\\

Upstream: https://github.com/github/copilot-sdk/issues/707


diberry:
Superseded by #699 — consolidated with #697. Both PRs touch squad-ci.yml for the same root cause; shipping as one clean PR.

PR #704, opened 2026-03-30
fix(cli): warn when squad watch receives unused message args (#703)

tamirdresher:
### What
Detects extra positional args passed to `squad watch` and prints a clear warning instead of silently ignoring them.

### Why
Users expect `squad watch --interval 5 "Nick, Run scheduled tasks"` to route the message to agent Nick on each poll cycle. But watch mode only runs Ralph polling loop — the quoted message is silently dropped. This caused confusion (reported by community user).

### How
After parsing `--interval`, check for remaining args. If found, print:
```
Warning: Watch mode does not route messages to agents. Ignoring: "Nick, Run scheduled tasks"
To address an agent directly, use an interactive session instead.
```

1 file changed: `packages/squad-cli/src/cli-entry.ts` (10 lines added)

Closes #703

### Testing
- [x] No functional change to watch behavior — warning only
- [x] Existing watch tests unaffected

### Docs
N/A

### Exports
N/A

### Breaking Changes
None

### Waivers
None

PR #858, opened 2026-04-05
fix(cli): warn when squad watch receives unused message args (#703)

### LINKED ISSUE #703: fix(cli): squad watch silently ignores quoted message
## Problem

`squad watch --interval 5 "Nick, Run scheduled tasks"` silently drops the quoted message. Ralph runs its normal polling loop with no indication the message was ignored.

Users expect the message to be routed to the named agent on each cycle, because direct addressing (`Nick, do something`) works in interactive sessions.

## Root Cause

In `cli-entry.ts` lines 322-329, only `--interval` is parsed. Extra args are silently ignored.

## Fix

Detect extra args and warn:
```
Warning: Watch mode does not route messages. Ignoring: "Nick, Run scheduled tasks"
To address Nick directly, use an interactive session.
```

## Files
- `packages/squad-cli/src/cli-entry.ts`
- `packages/squad-cli/src/cli/commands/watch.ts`

diberry:
### What

Detects extra positional args passed to `squad watch` and prints a clear warning instead of silently ignoring them.

### Why

Users expect `squad watch --interval 5 "Nick, Run scheduled tasks"` to route the message to agent Nick on each poll cycle. But watch mode only runs Ralph polling loop — the quoted message is silently dropped. This caused confusion (reported by community user).

### How

Makes the existing positional-arg detection (previously verbose-only) always warn with a clear UX message:

```
⚠️  Watch mode does not route messages to agents. Ignoring: "Nick, Run scheduled tasks"
   To address an agent directly, use an interactive session instead.
```

### Files Changed
- `packages/squad-cli/src/cli-entry.ts` — warning always shown (was verbose-only)
- `CHANGELOG.md` — changelog entry
- `docs/src/content/docs/features/ralph.md` — caution admonition
- `docs/src/content/docs/reference/cli.md` — CLI reference note
- `.changeset/watch-message-warning.md` — changeset

Closes #703

Based on PR #704 by @tamirdresher

Co-authored-by: tamirdresher
Co-authored-by: Copilot

### DIFF
--- a/.changeset/watch-message-warning.md
+++ b/.changeset/watch-message-warning.md
@@ -1,0 +1,5 @@
+---
+'@bradygaster/squad-cli': patch
+---
+
+Warn when squad watch receives unused positional arguments instead of silently ignoring them
--- a/docs/src/content/docs/features/ralph.md
+++ b/docs/src/content/docs/features/ralph.md
@@ -208,6 +208,21 @@ This runs as a standalone local process (not inside Copilot) that:
 - Assigns @copilot to `squad:copilot` issues (if auto-assign is enabled)
 - Runs until Ctrl+C
 
+> ⚠️ **Watch mode does not route messages**
+>
+> `squad watch` is a **triage polling loop**, not a message router. Extra arguments like agent names or messages are ignored:
+>
+> ```bash
+> # ❌ This does NOT route to Nick — the message is ignored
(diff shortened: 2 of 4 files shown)


diberry:
Closing — superseded by #859 which covers the same cli-entry.ts changes comprehensively.

Issue #939, opened 2026-04-09
squad init: Squad init should not run git init in a sub-folder of a repository (or ever)


serbrech:
We use a monorepo.
I've tried to setup squad in the folder that my team works in.
`squad init` detected that we're not at the root, warned about it, and then ran `git init` in the folder

Expectation: 
don't run `git init`.

```
Copilot resolves .github/agents/ from the git root, not from here.
The Squad agent won't be visible to copilot in this folder.
 
→ Running git init to create a repo boundary...
✓ Initialized git repo at /home/<user>/code/<repo-root>/<folder>
```

nirmitnd:
+1 to this. Even I use monorepo and the `git init` tripped me. The documentation also doesn't clearly mention that `git init` will be run.

tamirdresher:
Agreed — \git init\ should never run here. Filed **PR #944** with a fix: splits file placement (agent file at git root, .squad/ in subfolder), enables multiple squads per monorepo, and adds CWD-first resolution. See the PR for full details.

tamirdresher:
### Update: Workflow placement gap found during E2E testing

PR #944 fixes the core issue (no more `git init`, split file placement works), but a review found a gap: **GitHub Actions workflow files are placed in the subfolder in monorepo mode**, where GitHub can't find them.

Options being discussed on the PR:
1. Skip workflows entirely in monorepo-subfolder mode (multiple squads would conflict anyway)
2. Place at git root (but which squad's workflows win?)
3. Merge workflows intelligently (complex)

See [PR #944 comment](https://github.com/bradygaster/squad/pull/944#issuecomment-4213127988) for details. Would appreciate input on the preferred approach.

PR #944, opened 2026-04-09
fix: monorepo subfolder support — never run git init (#939)

tamirdresher:
## Problem

When `squad init` runs in a subfolder of a monorepo, it detects that CWD != git root and runs `git init` to create a nested repo boundary. This breaks monorepo workflows — users end up with git-in-git.

Reported in #939 with a +1 from another monorepo user.

## Solution

**Never run `git init`**. Instead, split file placement:

| File | Placed at | Why |
|------|-----------|-----|
| `.github/agents/squad.agent.md` | **Git root** | Copilot resolves custom agents from git root only |
| `.squad/` (team state) | **CWD** (subfolder) | Team state lives where the team works |

This enables multiple squads in a monorepo:

`
/monorepo/
  .github/agents/squad.agent.md   ← shared agent (at git root)
  team-alpha/.squad/               ← alpha's team state
  team-beta/.squad/                ← beta's team state
`

## Changes

1. **`packages/squad-cli/src/cli/core/init.ts`** — replaced `git init` block with monorepo-aware placement. Added `agentFileRoot` variable that points to git root when in subfolder.

2. **`packages/squad-sdk/src/config/init.ts`** — added `agentFileRoot` to `InitOptions` interface. Uses it for `.github/agents/` placement (falls back to `teamRoot`).

3. **Both `squad.agent.md.template` files** — updated Worktree Awareness section to check CWD first for `.squad/`, enabling monorepo subfolder resolution.

## How the agent finds the right `.squad/`

The coordinator resolves team root by checking **CWD first**:
1. Does CWD have `.squad/`? → use it (monorepo case)
2. Does git root have `.squad/`? → use it (normal case)
3. Fall back to main checkout discovery

Since Copilot CLI sets CWD to wherever you launched `copilot` from, `cd team-alpha && copilot` finds alpha's squad, `cd team-beta && copilot` finds beta's.

### DIFF
--- a/.changeset/monorepo-subfolder-support.md
+++ b/.changeset/monorepo-subfolder-support.md
@@ -1,0 +1,8 @@
+---
+"@bradygaster/squad-cli": patch
+"@bradygaster/squad-sdk": patch
+---
+
+fix: monorepo subfolder support — never run git init (#939)
+
+When `squad init` runs in a subfolder of a monorepo, place `.github/agents/squad.agent.md` at the git root and `.squad/` in the subfolder. Never run `git init`.
--- a/.github/agents/squad.agent.md
+++ b/.github/agents/squad.agent.md
@@ -625,15 +625,17 @@ Squad and all spawned agents may be running inside a **git worktree** rather tha
 
 **How the Coordinator resolves the team root (on every session start):**
 
-1. Run `git rev-parse --show-toplevel` to get the current worktree root.
-2. Check if `.squad/` exists at that root (fall back to `.ai-team/` for repos that haven't migrated yet).
+1. **Check CWD first** — does `.squad/` exist in the current working directory?
+   - **Yes** → Team root = CWD. This handles monorepos where `.squad/` lives in a subfolder.
+2. If not, run `git rev-parse --show-toplevel` to get the current worktree root.
+3. Check if `.squad/` exists at that root (fall back to `.ai-team/` for repos that haven't migrated yet).
    - **Yes** → use **worktree-local** strategy. Team root = current worktree root.
    - **No** → use **main-checkout** strategy. Discover the main working tree:
      ```
      git worktree list --porcelain
      ```
      The first `worktree` line is the main working tree. Team root = that path.
-3. The user may override the strategy at any time (e.g., *"use main checkout for team state"* or *"keep team state in this worktree"*).
+4. The user may override the strategy at any time (e.g., *"use main checkout for team state"* or *"keep team state in this worktree"*).
 
 **Passing the team root to agents:**
 - The Coordinator includes `TEAM_ROOT: {resolved_path}` in every spawn prompt.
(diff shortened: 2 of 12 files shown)


tamirdresher:
## 🔍 Review finding: Workflow placement gap in monorepo mode

After running full E2E tests, a review identified a gap:

**In monorepo mode, GitHub Actions workflow files (squad-heartbeat.yml, squad-triage.yml, etc.) are placed in the subfolder's .github/workflows/ — a location GitHub Actions completely ignores.** Only `<repo-root>/.github/workflows/` is processed by GitHub.

### Current behavior (with this PR)
- `squad.agent.md` → ✅ placed at git root (correct)
- `.squad/` → ✅ placed in subfolder (correct)
- `workflows/*.yml` → ❌ placed in subfolder (wrong — GitHub won't find them)

### Options

1. **Skip workflows in monorepo-subfolder mode** — simplest, and arguably correct since multiple squads in one monorepo would conflict on the same workflow files anyway. Each squad's labels/triage/heartbeat would step on each other.

2. **Place workflows at git root** — same approach as squad.agent.md. But raises the conflict question: which squad's triage workflow wins?

3. **Merge workflows intelligently** — combine label sets from all squads into one workflow. Most complex.

I'm leaning toward option 1 (skip) for now and tracking the multi-squad workflow story separately. Would love feedback from @bradygaster on the preferred approach.

serbrech:
> I'm leaning toward option 1 (skip) for now and tracking the multi-squad workflow story separately. Would love feedback from @bradygaster on the preferred approach.

I agree.
on the longer term, I lean towards 2. Workflows are a repo concept. Even if the workflow was to involve squad, the agent is defined at the gitroot, and the routing could still occur.

- can workflows be disabled (i.e ADO world...)?
- are the squad skills also using a relative path (should also be tied to the gitroot, as the skills go with the agent because copilot look at them at the root?)

Q: how do I move to this? just squad upgrade? it's fine if I need to move the files/folders manually.

tamirdresher:
## ✅ All review comments addressed (rebased + 4th commit)

Branch has been rebased onto latest \dev\ and all 12 review comments are fixed:

### Version bumps (3 comments)
- ✅ Reverted all \package.json\ versions from \

tamirdresher:
Hey @serbrech 

**Workflows:** Already skippable via `--no-workflows` flag. In monorepo mode this PR skips them automatically (since GitHub Actions ignores subfolder workflows anyway). Long-term we agree with option 2 — workflows at the git root with routing.

**Skills:** They resolve from CWD-relative `.copilot/skills/`, so each squad in a monorepo gets its own skill set. Works correctly because the agent template checks CWD first for `.squad/`.

**Migration:** For now it's 3 manual steps:
1. `rm -rf your-subfolder/.git` (delete the nested git)
2. `mv your-subfolder/.github/agents/squad.agent.md .github/agents/` (move agent to git root)
3. Keep `.squad/` in the subfolder as-is

`squad upgrade` doesn't handle monorepo relocation yet — only `squad init` is monorepo-aware in this PR. We'll track monorepo-aware upgrade as a follow-up issue. For now the manual steps above are the migration path.

PR #959, opened 2026-04-10
feat: shared squad — external state backend for multi-clone workflows

### LINKED ISSUE #958: feat: shared squad — external state backend for multi-clone workflows
## Problem
Some development workflows involve multiple clones of the same repository (worktrees, parallel clones, CI build environments,  #multiple dev machines). In these environments:
- Branch switches can destroy untracked files including `.squad/`
- Multiple checkouts of the same repo each need their own `squad init` - duplicating team identity
- Team identity, decisions, and agent knowledge can't be shared across clones
- `.squad/` committed to the repo isn't viable for large repos with strict policies on untracked content
- In very large monorepos, teams often work in isolated subtrees — committing `.squad/` at the repo root may not be practical or permitted

## Proposed Solution
Enable squad team state to live *outside* the repository — in a user-managed directory (git-backed squad repo or platform app data directory). Discovery happens via origin URL matching against a lightweight registry at `~/.squad/squad-repos.json`.

### Key design points
- **Zero repo writes** — shared mode writes nothing to the target repository
- **URL-based matching** — `git remote get-url origin` matched against registry patterns (GitHub, ADO, SSH URL forms all normalized)
- **Git-backed squad repos** — team state lives in its own repo, committable and shareable across machines
- **Per-clone runtime state** — clone-local ephemeral data stored under the platform's local app data directory (Windows/macOS/Linux)
- **Cross-platform** — `~/.squad/` pointer on all platforms; local data uses platform-appropriate paths

### New commands
- `squad init --shared` — create or connect to a shared squad
- `squad shared status|diagnose|list|doctor` — operational tooling
- `squad migrate --to shared` — move existing squad state externally

## Related
Partially addresses #943 (monorepo/persistence concerns)

akubly:
### What
Enable sharing a single squad team across multiple repository clones by storing team state externally in a git-backed squad repo. Adds a resolution chain that discovers shared state via `~/.squad/squad-repos.json` pointer files, per-clone runtime state isolation, and concurrent-safe Scribe merge via claim protocol.

### Why
Large monorepo workflows (e.g. repos with branch switching that destroys untracked files, multiple checkouts for parallel work, worktrees) lose `.squad/` state on branch switches. Committing `.squad/` to the repo isn't practical when the team is local. This feature externalizes squad state so it survives branch operations and can be shared across clones, machines, and worktrees.

Closes #958

### How
**Resolution chain** (priority order):
1. Local `.squad/` directory (unchanged, always wins)
2. `SQUAD_REPO_KEY` env -> `lookupByKeyAcrossRepos()` (checks git-backed pointers first, falls back to legacy)
3. URL-based discovery via origin remote -> `lookupByUrlAcrossRepos()` (same pointer-first strategy)
4. Legacy fallback
5. Init Mode (no squad found)

**Key components:**
- `shared-squad.ts` -- Registry CRUD, URL matching, git-backed repo pointers, path traversal validation
- `clone-state.ts` -- Per-clone runtime state (projectDir), generic leaf name handling to avoid collisions
- `scribe-merge.ts` -- Claim protocol for concurrent inbox merging (atomic claim files prevent last-writer-wins)
- `resolution-base.ts` -- Shared primitives extracted to break circular dependency between resolution.ts and shared-squad.ts
- CLI commands: `squad shared status|add-url|list|doctor|diagnose`
- `squad init --shared` and `squad migrate --to shared`
- `.squad/` validation: walk-up now requires `team.md` or `agents/` to avoid false positive on config directories

**Security hardening:**
- Path traversal validation on repo keys (F1)
- Agent name injection prevention in journal filenames (F5)
- Symlink/junction redirect detection via realpathSync (F7)
- Origin-only remote reading (F4)

---

### Quick Check
- [x] If SDK/CLI source files changed: completed the applicable Changeset step below

#### Branch and Commit
- [x] Branch created from dev (not main)
- [x] Branch is up to date with dev
- [x] Verified diff contains only intended changes (36 files)
- [ ] PR is not in draft mode (mark ready when checks pass)
- [x] Commit history is clean (single squashed commit)

#### Build and Test
- [x] npm run build passes (pre-existing node-pty error only)
- [x] npm test passes -- 259/260 feature tests pass, 0 regressions
- [x] npm run lint passes
- [x] npm run lint:eslint passes (0 errors)
- [x] Test output summary included below (36+ files changed)

#### Test Output Summary

Feature tests (11 files, 260 tests): 259 passed, 1 skipped, 0 failed.
Full suite (220 files): 214 passed, 5 failed (pre-existing), 1 skipped.
Pre-existing failures: docs-build (Astro), cli-packaging-smoke (node-pty), 3 vitest timeouts.

#### Changeset
- [x] Changeset added via npx changeset add (minor for both squad-sdk and squad-cli)

#### Docs
- N/A -- Design proposal at docs/proposals/ to be added in follow-up

#### Exports
- [x] package.json subpath exports updated (new modules: shared-squad, clone-state, scribe-merge, resolution-base)

---

### Breaking Changes
None -- new feature, existing resolution behavior preserved via fallback chain.

### Waivers
None required.

### DIFF
--- a/packages/squad-cli/src/cli/commands/migrate.ts
+++ b/packages/squad-cli/src/cli/commands/migrate.ts
@@ -4,7 +4,15 @@
  */
 
 import path from 'node:path';
-import { FSStorageProvider } from '@bradygaster/squad-sdk';
+import fs from 'node:fs';
+import { execSync } from 'node:child_process';
+import {
+  FSStorageProvider,
+  createSharedSquad,
+  normalizeRemoteUrl,
+  getRemoteUrl,
+  validateRepoKey,
+} from '@bradygaster/squad-sdk';
 
 const storage = new FSStorageProvider();
 import { success, warn, dim, bold, BOLD, RESET, DIM } from '../core/output.js';
@@ -18,9 +26,11 @@ import type {
 } from '@bradygaster/squad-sdk';
 
 export interface MigrateOptions {
-  to?: 'sdk' | 'markdown';
+  to?: 'sdk' | 'markdown' | 'shared';
   from?: 'ai-team';
   dryRun?: boolean;
+  key?: string;
+  keepLocal?: boolean;
 }
 
 interface ParsedTeam {
@@ -398,6 +408,230 @@ export async function runMigrate(cwd: string, options: MigrateOptions): Promise<
     return;
   }
   
+  // Helper: recursively copy a directory using StorageProvider
+  function copyDirRecursive(src: string, dest: string): void {
+    storage.mkdirSync(dest, { recursive: true });
+    const entries = storage.listSync?.(src) ?? [];
+    for (const entry of entries) {
+      const srcPath = path.join(src, entry);
+      const destPath = path.join(dest, entry);
+      if (storage.isDirectorySync(srcPath)) {
+        copyDirRecursive(srcPath, destPath);
+      } else {
+        const content = storage.readSync(srcPath);
+        if (content != null) {
+          storage.writeSync(destPath, content);
+        }
+      }
+    }
+  }
+
+  // Handle --to shared (local .squad/ → shared mode)
+  if (options.to === 'shared') {
+    if (mode === 'none') {
+      fatal('No squad found. Run `squad init` first.');
+    }
+    if (mode === 'legacy') {
+      fatal('Found .ai-team/ directory. Run `squad migrate --from ai-team` first.');
+    }
+
+    const squadDir = path.join(cwd, '.squad');
+    if (!storage.existsSync(squadDir)) {
+      fatal('No .squad/ directory found.');
+    }
+
+    // Determine repo key
+    let key = options.key;
+    let urlPatterns: string[] = [];
+
+    const remoteUrl = getRemoteUrl(cwd);
+    if (!key) {
+      if (!remoteUrl) {
+        fatal(
+          'Cannot auto-detect repo key: no git remote "origin" found.\n' +
+          '       Use --key <owner/repo> to specify the key explicitly.',
+        );
+      }
+      const normalized = normalizeRemoteUrl(remoteUrl);
+      if (normalized.provider === 'unknown') {
+        fatal(
+          `Could not derive a supported repo key from origin URL.\n` +
+          `       Remote: ${remoteUrl}\n` +
+          `       Use --key <owner/repo> to specify the key explicitly.`,
+        );
+      }
+      key = normalized.key;
+      urlPatterns = [normalized.normalizedUrl];
+    } else {
+      if (remoteUrl) {
+        const normalized = normalizeRemoteUrl(remoteUrl);
+        urlPatterns = [normalized.normalizedUrl];
+      }
+    }
+
+    try {
+      validateRepoKey(key);
+    } catch (err) {
+      fatal((err as Error).message);
+    }
+
+    console.log(`\n${BOLD}Squad Migrate${RESET} — local .squad/ → shared\n`);
+    console.log(`📦 Migrating local squad to shared...`);
+    console.log(`   Source: ${squadDir}`);
+
+    // Create shared squad
+    let teamDir: string;
+    try {
+      teamDir = createSharedSquad(key, urlPatterns);
+    } catch (err) {
+      fatal((err as Error).message);
+    }
+
+    console.log(`   Target: ${teamDir}`);
+    console.log('');
+
+    // Copy team-state directories and files
+    const teamDirs = ['agents', 'casting', 'skills', 'decisions', 'decisions/inbox'];
+    for (const dir of teamDirs) {
+      const srcDir = path.join(squadDir, dir);
+      const destDir = path.join(teamDir, dir);
+      if (storage.existsSync(srcDir)) {
+        copyDirRecursive(srcDir, destDir);
+        success(`Copying: ${dir}/`);
+      }
+    }
+
+    const teamFiles = ['team.md', 'routing.md', 'decisions.md'];
+    for (const file of teamFiles) {
+      const srcFile = path.join(squadDir, file);
+      const destFile = path.join(teamDir, file);
+      if (storage.existsSync(srcFile)) {
+        const content = storage.readSync(srcFile) ?? '';
+        storage.writeSync(destFile, content);
+        success(`Copying: ${file}`);
+      }
+    }
+
+    // Copy .github/agents/ to .github-template/agents/ in the shared squad
+    // so future `squad init --shared` connections can source the agent file
+    const githubAgentsDir = path.join(cwd, '.github', 'agents');
+    if (storage.existsSync(githubAgentsDir)) {
+      const templateAgentsDir = path.join(teamDir, '.github-template', 'agents');
+      copyDirRecursive(githubAgentsDir, templateAgentsDir);
+      success(`Copying: .github/agents/ → .github-template/agents/`);
+    }
+
+    if (urlPatterns.length > 0) {
+      console.log(`   Registered URL pattern: ${urlPatterns[0]}`);
+    }
+
+    console.log('');
+    console.log(`✅ Migrated to shared squad "${key}"`);
+
+    // Cleanup local files after successful migration
+    if (!options.keepLocal) {
+      console.log('');
+      console.log(`🧹 Cleaning up local squad files...`);
+
+      // Clean up .squad/ directory
+      if (storage.existsSync(squadDir)) {
+        // Check if .squad/ contains any git-tracked files
+        let hasTrackedFiles = false;
+        try {
+          const tracked = execSync('git ls-files .squad/', { cwd, encoding: 'utf-8' }).trim();
+          hasTrackedFiles = tracked.length > 0;
+        } catch {
+          // git ls-files failed — treat as untracked to be safe
+        }
+
+        if (hasTrackedFiles) {
+          warn(`  .squad/ contains git-tracked files — left in place`);
+          console.log(`         ${DIM}Run \`git rm -r .squad/\` to remove tracked files${RESET}`);
+        } else {
+          try {
+            fs.rmSync(squadDir, { recursive: true, force: true });
+            success('  Removed .squad/');
+          } catch (err) {
+            warn(`  Could not remove .squad/: ${(err as Error).message}`);
+          }
+        }
+      }
+
+      // Clean up .github/agents/squad.agent.md (only if untracked)
+      const agentFile = path.join(cwd, '.github', 'agents', 'squad.agent.md');
+      if (storage.existsSync(agentFile)) {
+        let isTracked = false;
+        try {
+          const tracked = execSync('git ls-files .github/agents/squad.agent.md', { cwd, encoding: 'utf-8' }).trim();
+          isTracked = tracked.length > 0;
+        } catch { /* ignore */ }
+
+        if (isTracked) {
+          warn(`  .github/agents/squad.agent.md is git-tracked — left in place`);
+        } else {
+          try {
+            fs.unlinkSync(agentFile);
+            success('  Removed .github/agents/squad.agent.md');
+            // Clean up empty .github/agents/ dir
+            const agentsDir = path.join(cwd, '.github', 'agents');
+            try {
+              const remaining = fs.readdirSync(agentsDir);
+              if (remaining.length === 0) fs.rmdirSync(agentsDir);
+            } catch { /* ignore */ }
+          } catch (err) {
+            warn(`  Could not remove agent file: ${(err as Error).message}`);
+          }
+        }
+      }
+
+      // Clean up .gitattributes if it was squad-generated and untracked
+      const gitattributes = path.join(cwd, '.gitattributes');
+      if (storage.existsSync(gitattributes)) {
+        let isTracked = false;
+        try {
+          const tracked = execSync('git ls-files .gitattributes', { cwd, encoding: 'utf-8' }).trim();
+          isTracked = tracked.length > 0;
+        } catch { /* ignore */ }
+
+        if (!isTracked) {
+          // Only remove if it looks squad-generated (contains merge=union for .squad/)
+          const content = storage.readSync(gitattributes) ?? '';
+          if (content.includes('.squad/') && content.includes('merge=union')) {
+            try {
+              fs.unlinkSync(gitattributes);
+              success('  Removed .gitattributes (squad-generated)');
+            } catch { /* ignore */ }
+          }
+        }
+      }
+
+      // Clean up .copilot/ if untracked
+      const copilotDir = path.join(cwd, '.copilot');
+      if (storage.existsSync(copilotDir)) {
+        let isTracked = false;
+        try {
+          const tracked = execSync('git ls-files .copilot/', { cwd, encoding: 'utf-8' }).trim();
+          isTracked = tracked.length > 0;
+        } catch { /* ignore */ }
+
+        if (!isTracked) {
+          try {
+            fs.rmSync(copilotDir, { recursive: true, force: true });
+            success('  Removed .copilot/');
+          } catch { /* ignore */ }
+        }
+      }
+
+      console.log('');
+      console.log(`   ${DIM}Use --keep-local to skip cleanup next time.${RESET}`);
+    } else {
+      console.log('');
+      console.log(`   ${DIM}Local files left in place (--keep-local).${RESET}`);
+      console.log(`   ${DIM}Run \`git clean -xdf .squad .github/agents .gitattributes .copilot\` to remove manually.${RESET}`);
+    }
+    return;
+  }
+
   // Handle --to markdown (reverse migration)
   if (options.to === 'markdown') {
     if (mode !== 'sdk') {
--- a/packages/squad-sdk/src/clone-state.ts
+++ b/packages/squad-sdk/src/clone-state.ts
@@ -1,0 +1,269 @@
+/**
+ * Clone-local runtime state resolution.
+ *
+ * Derives and manages per-clone state directories stored outside the repo
+ * working tree, under the platform-specific LOCAL app data directory.
+ *
+ * Layout: {localBase}/squad/repos/{repo-key}/clones/{leaf-name}/
+ *
+ * Uses `validateRepoKey()` from shared-squad.ts for consistent validation.
+ *
+ * @module clone-state
+ */
+
+import path from 'node:path';
+import os from 'node:os';
+import { FSStorageProvider } from './storage/fs-storage-provider.js';
+import { validateRepoKey } from './shared-squad.js';
+import { CASE_INSENSITIVE } from './resolution-base.js';
+
+const storage = new FSStorageProvider();
+
+/**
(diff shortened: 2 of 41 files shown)


packages/squad-sdk/src/resolution.ts
```diff
@@ -9,13 +9,24 @@
...
  * @module resolution
  */
 
 import path from 'node:path';
 import os from 'node:os';
 import crypto from 'node:crypto';
+import { realpathSync } from 'node:fs';
 import { FSStorageProvider } from './storage/fs-storage-provider.js';
+import { SquadError, ErrorSeverity, ErrorCategory } from './adapter/errors.js';
+import { resolveSharedSquad, loadRepoRegistry, validateRepoKey } from './shared-squad.js';
```
akubly: **Must-fix: Circular import** — `resolution.ts` imports from `shared-squad.ts`, and `shared-squad.ts` imports back from `resolution.ts`. The comment above says it's safe because of hoisted function declarations, and that's true _today_ — but it's fragile. Adding a top-level `const` that depends on a cross-module export would break silently at runtime.

Let's extract the shared primitives (`resolveGlobalSquadPath`, `resolvePersonalSquadDir`, `ResolvedSquadPaths` type) into a `resolution-base.ts`. Both files import from it, no cycle.

packages/squad-sdk/src/scribe-merge.ts
```diff
@@ -0,0 +1,375 @@
...
+ * Implements the Scribe Claim Protocol from the shared-squad-across-clones
+ * design: atomic rename from inbox/ → processing/, merge into canonical
+ * file with content-hash deduplication, crash recovery for stale
+ * processing/ entries.
+ *
+ * @module scribe-merge
+ */
+
+import { createHash, randomBytes } from 'node:crypto';
+import {
```
akubly: **Must-fix: Bypasses StorageProvider** — This is the only SDK module that imports raw `node:fs` functions (9 of them) instead of going through `StorageProvider`. Every other module uses the abstraction. Makes this untestable with mock storage and breaks the pattern.

Should accept `StorageProvider` as a parameter. For the `renameSync` (needed for the atomic claim), either extend `StorageProvider` with a `rename()` method, or inject a narrow `{ rename: (src, dest) => void }` dependency just for that.

packages/squad-sdk/src/clone-state.ts
```diff
@@ -0,0 +1,288 @@
...
+  ) {
+    resolved = resolved.toLowerCase();
+  }
+  return resolved;
+}
+
+/**
+ * Validate repo key segments: must be non-empty, lowercase, no path traversal.
+ */
+function validateRepoKey(repoKey: string): void {
```
akubly: **Must-fix: Duplicate validation** — There's a thorough 40-line `validateRepoKey()` in `shared-squad.ts` that checks null bytes, max-length, Windows-illegal chars, and more. This private copy is only 15 lines and misses several of those checks.

Just delete this and import the canonical one from `shared-squad.ts`.

packages/squad-sdk/src/tools/index.ts
```diff
@@ -280,16 +292,13 @@ export class ToolRegistry {
           return { textResultForLlm: 'Invalid author name: must contain only letters, numbers, hyphens, and underscores', resultType: 'failure', error: 'Invalid author' };
         }
         try {
-          const inboxDir = path.join(this.squadRoot, 'decisions', 'inbox');
+          const inboxDir = path.join(this.resolvedPaths.teamDir, 'decisions', 'inbox');
```
akubly, line 295: **Must-fix: No path validation before write** — This now writes to `resolvedPaths.teamDir` instead of `this.squadRoot`. If a registry entry were ever compromised (e.g. pointing outside the global squad root via symlink), this could write to arbitrary locations.

Add a `validateWritePath(inboxDir, globalSquadRoot)` check before writing. Defense-in-depth — `createSharedSquad()` already validates paths, but the tool handlers should too.

packages/squad-sdk/src/shared-squad.ts
```diff
@@ -0,0 +1,626 @@
...
+ * Uses `fs.realpathSync()` on the nearest existing ancestor of `resolvedPath`
+ * and on `expectedRoot` to catch symlink/junction redirect attacks. This is
+ * safe to call even when the target file doesn't exist yet — it walks up to
+ * the nearest existing ancestor directory.
+ *
+ * @param resolvedPath  - The target path to validate (may not exist yet).
+ * @param expectedRoot  - The directory the path must stay inside (must exist).
+ * @throws {Error} If the path escapes the expected root.
+ */
+export function validateWritePath(resolvedPath: string, expectedRoot: string): void {
```
akubly, line 114: **Must-fix: Missing directory check** — If `expectedRoot` is accidentally a file path instead of a directory, `realpathSync` resolves it and `pathStartsWith` might incorrectly pass. Add a directory check after resolving:
```typescript
if (!storage.isDirectorySync(resolvedRoot)) {
  throw new Error(`Write path validation failed: expected root is not a directory`);
}
```

packages/squad-sdk/src/resolution.ts
```diff
@@ -116,7 +132,13 @@ export function resolveSquad(startDir?: string): string | null {
     const candidate = path.join(current, '.squad');
 
     if (storage.existsSync(candidate) && storage.isDirectorySync(candidate)) {
-      return candidate;
+      // Validate this is a real squad team root, not just a config directory
+      // (e.g. ~/.squad/ which only contains squad-repos.json pointer files).
+      const hasTeam = storage.existsSync(path.join(candidate, 'team.md'));
```
akubly, line 101: **Should-fix: Behavior change could break edge cases** — Adding the `team.md || agents/` requirement changes `resolveSquad()`'s contract. A freshly-created `.squad/` mid-scaffolding or a manual remote-mode setup with only `config.json` would stop being found.

Safer to also accept `config.json`: `hasTeam || hasAgents || hasConfig`. Or move this filtering into `resolveSquadPaths()` where we know the context (shared vs. local).

packages/squad-sdk/src/shared-squad.ts
```diff
@@ -0,0 +1,626 @@
...
+import { resolveGlobalSquadPath, resolvePersonalSquadDir } from './resolution.js';
+import type { ResolvedSquadPaths } from './resolution.js';
+import { normalizeRemoteUrl, getRemoteUrl } from './platform/detect.js';
+import { resolveCloneStateDir } from './clone-state.js';
+
+const storage = new FSStorageProvider();
+
+// Case-insensitive path comparison on Windows and macOS (default HFS+/APFS).
+// macOS CAN be case-sensitive on rare APFS configurations; set SQUAD_CASE_SENSITIVE=1 to override.
+const CASE_INSENSITIVE =
```
akubly: **Should-fix: Duplicated constant** — `CASE_INSENSITIVE` is defined with the same logic in three modules: here, `resolution.ts`, and `clone-state.ts` (via `normalizePath`). Extract it to a shared utility (`platform/detect.ts` would be natural) so it doesn't drift.

packages/squad-cli/src/cli/commands/migrate.ts
```diff
@@ -21,6 +29,8 @@ export interface MigrateOptions {
   to?: 'sdk' | 'markdown';
```
akubly: **Should-fix: Type doesn't match runtime** — This still says `'sdk' | 'markdown'` but `cli-entry.ts` passes `'shared'` into it. The cast bypasses TypeScript's safety. Update to `'sdk' | 'markdown' | 'shared'`.

packages/squad-sdk/src/resolution.ts
```diff
@@ -256,39 +283,173 @@ export function isConsultMode(config: SquadDirConfig | null): boolean {
...
+ *
+ * @throws {SquadError} If `%APPDATA%` (or override) is unreachable (F11).
+ */
+function resolveSharedMode(startDir: string): ResolvedSquadPaths | null {
+  const repoRoot = findGitRoot(startDir);
+  if (!repoRoot) return null;
+
+  // SQUAD_APPDATA_OVERRIDE: log once when entering shared discovery
+  if (process.env['SQUAD_APPDATA_OVERRIDE']) {
+    console.warn(
```
akubly, line 342: **Should-fix: Side effect in resolution function** — `console.warn()` in a function that should be pure. This will fire every time `resolveSharedMode()` is called, which could be multiple times per CLI invocation. Return the override status in the result instead, and let the caller decide whether to log.

packages/squad-sdk/src/index.ts
```diff
@@ -10,8 +10,39 @@ const pkg = require('../package.json');
 export const VERSION: string = pkg.version;
 
 // Export public API
-export { resolveSquad, resolveGlobalSquadPath, resolvePersonalSquadDir, ensurePersonalSquadDir, ensureSquadPath, ensureSquadPathTriple, loadDirConfig, isConsultMode, scratchDir, scratchFile, deriveProjectKey, resolveExternalStateDir } from './resolution.js';
+export { resolveSquad, resolveGlobalSquadPath, resolvePersonalSquadDir, ensurePersonalSquadDir, ensureSquadPath, ensureSquadPathTriple, loadDirConfig, isConsultMode, scratchDir, scratchFile, deriveProjectKey, resolveExternalStateDir, resolveSquadPaths } from './resolution.js';
+export { ensureCloneState, resolveCloneStateDir } from './clone-state.js';
+export { normalizeRemoteUrl, getRemoteUrl } from './platform/detect.js';
+export type { NormalizedRemote } from './platform/detect.js';
```
akubly, line 16: **Should-fix: Large export surface** — 30+ new symbols exported here, including internal utilities like `sanitizeJournalFilenameComponent`, `validateWritePath`, `recoverStaleProcessing`. Not all of these should be public API — they increase the surface that has to be maintained under semver.

Worth auditing which are genuinely needed by CLI consumers vs. which are internal. Could use a `shared-squad/internal` subpath export for the non-public ones.

akubly (commented):
Code review panel findings.

packages/squad-sdk/src/resolution.ts
akubly: Fixed in 0ec84328 — extracted 
esolveGlobalSquadPath, 
esolvePersonalSquadDir, ResolvedSquadPaths, CASE_INSENSITIVE, and pathStartsWith into 
esolution-base.ts. Both 
esolution.ts and shared-squad.ts now import from the base module. Circular import eliminated, stale comment removed.

packages/squad-sdk/src/scribe-merge.ts
akubly: Fixed in 32a271cf — all 9 raw 
ode:fs imports replaced with StorageProvider methods. Added optional storage parameter (defaults to FSStorageProvider) to all public functions. 
enameSync is now storage.renameSync() after extending the provider. All 21 scribe-merge tests pass unchanged.

packages/squad-sdk/src/clone-state.ts
akubly: Fixed in 0ec84328 — deleted the private 15-line copy and imported the canonical alidateRepoKey from shared-squad.ts.

packages/squad-sdk/src/tools/index.ts
akubly: Fixed in 32a271cf — added alidateWritePath checks in both squad_decide and squad_memory handlers. Validation runs in shared mode only (local mode's 	eamDir is the trusted in-repo .squad/).

packages/squad-sdk/src/shared-squad.ts
akubly: Fixed in 0ec84328 — added storage.isDirectorySync(resolvedRoot) check after 
ealpathSync resolution. Throws descriptive error if the expected root is not a directory.

packages/squad-sdk/src/resolution.ts
akubly: Fixed in 0ec84328 — added config.json to the validation: hasTeam || hasAgents || hasConfig. Applied in both 
esolveSquad() and indSquadDir().

packages/squad-sdk/src/shared-squad.ts
akubly: Fixed in 0ec84328 — CASE_INSENSITIVE and pathStartsWith extracted to 
esolution-base.ts as the single source of truth. All three modules now import from there.

packages/squad-cli/src/cli/commands/migrate.ts
akubly: Fixed in 32a271cf — updated type to `'sdk' | 'markdown' | 'shared'`.

packages/squad-sdk/src/resolution.ts
akubly: Fixed in 0ec84328 — added a module-level `_appdataOverrideWarned` flag so the warning fires at most once per process.

packages/squad-sdk/src/index.ts
akubly: Fixed in 32a271cf — added `@internal` JSDoc tags to alidateWritePath, sanitizeJournalFilenameComponent, loadSquadRepoPointers, mergeInbox, and 
ecoverStaleProcessing. They remain exported (CLI needs them) but are marked as not part of the public API surface.

packages/squad-sdk/src/shared-squad.ts
```diff
@@ -0,0 +1,617 @@
...
+  if (WINDOWS_ILLEGAL_CHARS.test(key)) {
+    throw new Error(
+      `Invalid repo key "${key}": contains illegal characters (< > : " | ? * \\)`
+    );
+  }
+
+  // Path traversal — check before splitting since `..` could appear in any position
+  if (key.includes('..')) {
+    throw new Error(`Invalid repo key "${key}": path traversal (..) rejected`);
+  }
```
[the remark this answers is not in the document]
akubly: Good catch — fixed in 1f37d977. Changed from substring `key.includes('..')` to segment-aware `segments.some(s => s === '.' || s === '..')`. Repo names like `repo..v2` are now correctly accepted.

packages/squad-sdk/src/shared-squad.ts
```diff
@@ -0,0 +1,617 @@
...
+ * - F7 (IMPORTANT): Symlink/junction redirect attacks on write paths
+ *
+ * @module shared-squad
+ */
+
+import path from 'node:path';
+import os from 'node:os';
+import { realpathSync } from 'node:fs';
+import { FSStorageProvider } from './storage/fs-storage-provider.js';
+import { resolveGlobalSquadPath, resolvePersonalSquadDir, CASE_INSENSITIVE, pathStartsWith } from './resolution-base.js';
```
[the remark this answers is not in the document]
akubly: Fixed in 1f37d977 — removed the unused `CASE_INSENSITIVE` import.

packages/squad-sdk/src/tools/index.ts
```diff
@@ -386,32 +403,23 @@ export class ToolRegistry {
...
-          const sectionIndex = content.indexOf(sectionHeader);
-          if (sectionIndex !== -1) {
-            // Find next section or end of file
-            const nextSectionIndex = content.indexOf('\n## ', sectionIndex + sectionHeader.length);
-            const insertIndex = nextSectionIndex === -1 ? content.length : nextSectionIndex;
-            content = content.slice(0, insertIndex) + entry + content.slice(insertIndex);
-          } else {
-            // Section doesn't exist, append at end
-            content += `\n${sectionHeader}\n${entry}`;
+          const inboxDir = path.join(this.resolvedPaths.teamDir, 'agents', args.agent, 'history', 'inbox');
```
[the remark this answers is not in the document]
akubly: Fixed in 1f37d977 — history existence check now uses `this.resolvedPaths.teamDir` instead of `this.squadRoot`.

packages/squad-sdk/src/scribe-merge.ts
```diff
@@ -0,0 +1,372 @@
...
+ */
+function splitEntries(content: string): string[] {
+  if (!content.trim()) return [];
+  const blocks: string[] = [];
+  // Split on lines that start with ### (keeping the delimiter)
+  const parts = content.split(/^(?=### )/m);
+  for (const part of parts) {
+    const trimmed = part.trim();
+    if (trimmed) blocks.push(trimmed);
+  }
```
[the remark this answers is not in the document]
akubly: Fixed in 1f37d977 — rewrote `splitEntries()` to buffer `## ` section headers and attach them to the following `### ` entry block. Format stays consistent between inbox files and canonical history.md.

packages/squad-cli/src/cli/commands/migrate.ts
```diff
@@ -398,6 +408,234 @@ export async function runMigrate(cwd: string, options: MigrateOptions): Promise<
...
+
+    // Cleanup local files after successful migration
+    if (!options.keepLocal) {
+      console.log('');
+      console.log(`🧹 Cleaning up local squad files...`);
+
+      // Determine which local squad files are tracked vs untracked
+      const localSquadFiles = ['.squad', '.github/agents/squad.agent.md'];
+      const localMetaFiles = ['.gitattributes', '.copilot'];
+
```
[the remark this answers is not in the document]
akubly: Fixed in 1f37d977 — removed both unused variables.

packages/squad-sdk/src/tools/index.ts
```diff
@@ -280,16 +293,19 @@ export class ToolRegistry {
...
         }
         try {
-          const inboxDir = path.join(this.squadRoot, 'decisions', 'inbox');
+          const inboxDir = path.join(this.resolvedPaths.teamDir, 'decisions', 'inbox');
+
+          // Validate write target stays within expected root (shared mode only —
+          // in local mode teamDir is the trusted in-repo .squad/ directory)
+          if (this.resolvedPaths.mode === 'shared') {
+            validateWritePath(inboxDir, resolveGlobalSquadPath());
+          }
```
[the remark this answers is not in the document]
akubly: Fixed in 1f37d977 — now validates against `this.resolvedPaths.teamDir` instead of `resolveGlobalSquadPath()`. Works for both global-dir and git-backed shared squads.

packages/squad-sdk/src/tools/index.ts
```diff
@@ -386,32 +403,23 @@ export class ToolRegistry {
...
-            content = content.slice(0, insertIndex) + entry + content.slice(insertIndex);
-          } else {
-            // Section doesn't exist, append at end
-            content += `\n${sectionHeader}\n${entry}`;
+          const inboxDir = path.join(this.resolvedPaths.teamDir, 'agents', args.agent, 'history', 'inbox');
+
+          // Validate write target stays within expected root (shared mode only —
+          // in local mode teamDir is the trusted in-repo .squad/ directory)
+          if (this.resolvedPaths.mode === 'shared') {
+            validateWritePath(inboxDir, resolveGlobalSquadPath());
```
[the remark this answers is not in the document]
akubly: Fixed in 1f37d977 — same change as comment 6, validates against `this.resolvedPaths.teamDir`. Also removed the now-unused `resolveGlobalSquadPath` import.

PR #988, opened 2026-04-16
fix: replace obsolete --message flag with -p in watch and loop

### LINKED ISSUE #874: loop uses obsolete gh copilot --message flag
## Bug

`loop` and related watch/execute paths invoke `gh copilot --message ...`, but the currently installed GitHub CLI Copilot integration on this machine accepts `-p` rather than `--message`.

This causes the loop to fail immediately even when the squad setup itself is healthy.

## Environment

- `@bradygaster/squad-cli`: `0.9.4-insider.1`
- OS: WSL2 Ubuntu / Linux
- Node: `v22.22.1`
- npm: `10.9.4`
- GitHub CLI: authenticated and working
- `gh copilot --help` shows prompt usage as `gh copilot -p "..."`

## Repro

1. Create a valid `loop.md` with `configured: true`
2. Run:

```bash
npm exec squad -- loop
```

## Actual

The first round starts and then fails with:

```text
error: unknown option '--message'
Try 'copilot --help' for more information.
```

The failing constructed command is effectively:

```bash
gh copilot --message "...loop prompt..."
```

## Expected

`loop` should invoke the installed `gh copilot` in a way that is accepted by the current CLI.

## Evidence

On this machine:

```bash
gh copilot --help
```

shows examples like:

```bash
gh copilot -p "Summarize this week's commits"
```

and does not document `--message`.

## Root Cause

The installed Squad CLI still hardcodes `--message` in multiple places, including:

- `dist/cli/commands/loop.js`
- `dist/cli/commands/watch/index.js`
- `dist/cli/commands/watch/capabilities/*.js`

Example from the installed code:

```js
const args = ['copilot', '--message', prompt];
```

## Local workaround

Replacing `--message` with `-p` in the installed CLI allowed the loop to get past the previous immediate failure and begin Round 1 successfully.

## Suggested fix

Update the CLI integration points to use the currently supported `gh copilot` prompt flag shape, or centralize the Copilot command construction behind a compatibility layer so this does not break again when GitHub CLI changes.

### LINKED ISSUE #980: Bug: watch --execute fails — gh copilot --message is not a valid command
## Summary

`squad watch --execute` has several bugs that prevent it from working. All are fixable. This issue documents root causes and tested fixes.

## 1. Wrong Copilot CLI invocation

`execute.js buildAgentCommand` uses `gh copilot --message` — neither `--message` nor `gh copilot` as a subcommand exist. The Copilot CLI binary is `copilot` (or `gh copilot --` to pass through), and the non-interactive flag is `-p`.

Additionally, without permission flags, copilot answers in text but **cannot do work** (no file I/O, no shell, no git).

```js
// Broken:
const args = ['copilot', '--message', prompt];
return { cmd: 'gh', args };

// Working:
const args = ['-p', prompt, '--yolo', '--no-ask-user', '--silent'];
return { cmd: 'copilot', args };
```

Use `copilot` directly, not `gh copilot` — on Windows `gh` is a `.cmd` batch file that opens a visible console window even with `windowsHide: true`.

## 2. Config file `execute: true` doesn't enable the capability

The CLI flag path (`--execute`) explicitly sets `capabilities['execute'] = true`. The config file path reads `execute: true` into `config.execute` but never maps it to `config.capabilities`. The preflight system only checks `config.capabilities[cap.name]` — so the capability silently doesn't load.

```js
// Fix in loadWatchConfig (config.js), after the merge block:
if (merged.execute && !merged.capabilities['execute']) {
    merged.capabilities['execute'] = true;
}
```

## 3. Single prompt for all issues blocks Ralph

`executeAll` sends every eligible issue in one copilot prompt and waits synchronously (up to 30min timeout). Ralph can't poll or triage during this time.

**Fix:** Spawn one detached copilot process per issue (max N per round). Use `spawn()` with `detached: true, stdio: 'ignore'` so Ralph returns immediately. Track dispatched issue numbers in a state file to avoid re-dispatching on the next poll.

## 4. Console windows on Windows (MCP servers via npx)

Copilot spawns MCP servers (Context7, WorkIQ, Playwright) via `npx`, which is a `.cmd` batch file on Windows. Every `.cmd` invocation opens a visible console window — 10-15 of them per copilot session, coming and going.

`windowsHide: true` on the parent spawn does not propagate to grandchild processes.

**Workaround:** Use `--config-dir` pointing to a headless MCP config where stdio servers use `node` directly instead of `npx`:

```json
{
  "mcpServers": {
    "Context7": {
      "type": "stdio",
      "command": "node",
      "args": ["/path/to/node_modules/@upstash/context7-mcp/dist/index.js"]
    }
  }
}
```

`node.exe` is a real executable — no console window. HTTP-based MCP servers (Microsoft Docs, FluentUI) don't spawn subprocesses and work as-is.

**Proper fix (in copilot CLI):** When running in non-interactive mode (`-p`), copilot should spawn MCP stdio subprocesses with `windowsHide: true` or the Windows `CREATE_NO_WINDOW` flag. Alternatively, provide a `--no-mcp` or `--headless` flag.

## 5. No worktree isolation

Multiple copilot sessions in the same working directory cause git conflicts. Each issue should get its own worktree:

```js
execFileSync('git', ['worktree', 'add', `../repo-${issue.number}`, '-b', `squad/${issue.number}-${slug}`, 'main']);
spawn('copilot', args, { cwd: worktreePath, detached: true, ... });
```

## Recommended `executeAll` architecture

```
Per round (non-blocking):
  1. Load state file → skip already-dispatched issues
  2. Pick up to 3 eligible issues
  3. For each:
     a. git worktree add ../repo-{N} -b squad/{N}-{slug} main
     b. spawn('copilot', ['-p', prompt, '--yolo', '--no-ask-user', '--silent',
              '--config-dir', '.squad/.copilot-headless'],
              { cwd: worktreePath, detached: true, stdio: 'ignore', windowsHide: true })
  4. Save dispatched issue numbers to state file
  5. Return immediately → Ralph keeps polling
```

## Environment

- Squad CLI: v0.9.4-insider.1
- OS: Windows 11
- Copilot CLI: latest (WinGet)

All fixes tested and working locally via monkey-patches to the installed squad CLI.

tamirdresher:
## Summary

Replace obsolete `--message` flag with `-p` across all watch capabilities and loop command. Also switch default invocation from `gh copilot` to `copilot` directly to avoid Windows console window issues.

## Root cause

The Copilot CLI uses `-p` for non-interactive prompts. `--message` was never a valid flag and causes immediate failure: `error: unknown option '--message'`.

Additionally, `gh` on Windows is a `.cmd` batch file that spawns a visible console window even with `windowsHide: true`. Using `copilot` directly avoids this.

## Changes

**8 source files fixed** (same pattern in each `buildAgentCommand` / `buildLoopAgentCommand`):
- `packages/squad-cli/src/cli/commands/watch/capabilities/execute.ts`
- `packages/squad-cli/src/cli/commands/watch/capabilities/decision-hygiene.ts`
- `packages/squad-cli/src/cli/commands/watch/capabilities/monitor-email.ts`
- `packages/squad-cli/src/cli/commands/watch/capabilities/monitor-teams.ts`
- `packages/squad-cli/src/cli/commands/watch/capabilities/retro.ts`
- `packages/squad-cli/src/cli/commands/watch/capabilities/wave-dispatch.ts`
- `packages/squad-cli/src/cli/commands/watch/index.ts`
- `packages/squad-cli/src/cli/commands/loop.ts`

**1 pre-existing fix** (unblocks build):
- `packages/squad-sdk/src/platform/comms-teams.ts` — `TOKEN_PATH` typo from #906

**1 test file updated:**
- `test/cli/watch-execute.test.ts` — assertions updated to expect `copilot` and `-p`

All 16 watch-execute tests pass.

Fixes #980
Closes #874

### DIFF
--- a/.changeset/fix-copilot-message-flag.md
+++ b/.changeset/fix-copilot-message-flag.md
@@ -1,0 +1,11 @@
+---
+'@bradygaster/squad-cli': patch
+'@bradygaster/squad-sdk': patch
+---
+
+Fix obsolete --message flag in watch and loop commands
+
+Replace `--message` with `-p` across all watch capabilities and loop command.
+Use `copilot` directly on Windows to avoid .cmd console window issues,
+fall back to `gh copilot` on other platforms. Fix TOKEN_PATH typo in
+comms-teams.ts (pre-existing bug from #906).
--- a/packages/squad-cli/src/cli/commands/loop.ts
+++ b/packages/squad-cli/src/cli/commands/loop.ts
@@ -136,13 +136,13 @@ function buildLoopAgentCommand(
 ): { cmd: string; args: string[] } {
   if (options.agentCmd) {
     const parts = options.agentCmd.trim().split(/\s+/);
-    return { cmd: parts[0]!, args: [...parts.slice(1), '--message', prompt] };
+    return { cmd: parts[0]!, args: [...parts.slice(1), '-p', prompt] };
   }
-  const args = ['copilot', '--message', prompt];
+  const args = ['-p', prompt];
   if (options.copilotFlags) {
     args.push(...options.copilotFlags.trim().split(/\s+/));
   }
-  return { cmd: 'gh', args };
+  return { cmd: 'copilot', args };
 }
 
 // ── Capability Phase Runner ──────────────────────────────────────
@@ -244,10 +244,10 @@ function createNoopAdapter(): ReturnType<typeof createPlatformAdapter> {
 
 // ── gh Copilot Preflight ─────────────────────────────────────────
 
-/** Verify `gh` CLI with copilot extension is available. */
-async function checkGhCopilot(): Promise<void> {
+/** Verify the copilot CLI is available. */
+async function checkCopilotCli(): Promise<void> {
   return new Promise<void>((resolve, reject) => {
-    execFile('gh', ['copilot', '--version'], (err) => {
+    execFile('copilot', ['--version'], (err) => {
       if (err) reject(err);
       else resolve();
     });
@@ -313,12 +313,12 @@ export async function runLoop(dest: string, options: LoopConfig): Promise<void> 
     fatal('timeout must be a positive number of minutes');
   }
 
-  // Preflight: verify gh copilot is available (skip if user overrides the agent command)
+  // Preflight: verify copilot CLI is available (skip if user overrides the agent command)
   if (!options.agentCmd) {
     try {
-      await checkGhCopilot();
+      await checkCopilotCli();
     } catch {
-      fatal('gh CLI with copilot extension required. Install from https://cli.github.com/ and run `gh extension install github/gh-copilot`');
+      fatal('Copilot CLI required. Install from https://cli.github.com/ and run `gh extension install github/gh-copilot`');
     }
   }
 
--- a/packages/squad-cli/src/cli/commands/watch/capabilities/decision-hygiene.ts
+++ b/packages/squad-cli/src/cli/commands/watch/capabilities/decision-hygiene.ts
@@ -12,11 +12,11 @@ const storage = new FSStorageProvider();
 function buildAgentCommand(prompt: string, context: WatchContext): { cmd: string; args: string[] } {
   if (context.agentCmd) {
     const parts = context.agentCmd.trim().split(/\s+/);
-    return { cmd: parts[0]!, args: [...parts.slice(1), '--message', prompt] };
+    return { cmd: parts[0]!, args: [...parts.slice(1), '-p', prompt] };
   }
-  const args = ['copilot', '--message', prompt];
+  const args = ['-p', prompt];
   if (context.copilotFlags) args.push(...context.copilotFlags.trim().split(/\s+/));
-  return { cmd: 'gh', args };
+  return { cmd: 'copilot', args };
 }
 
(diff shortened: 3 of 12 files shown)


diberry:
Hey @tamirdresher — nice fix. Here's a review:

## PR #988 Review

### 1. Is this a fix for the issue?

**Yes.** The PR correctly updates `--message` to `-p` across all 8 files that build Copilot CLI commands, and fixes the `TOKEN_PATH` reference that would fail on Windows. Tests are updated. The immediate bugs are resolved.

### 2. Is this a fix for the class of issue?

**Not yet.** The underlying pattern — the same command-building logic living in 8 separate places — is still there. This PR applies the right value everywhere, which is the right first step. A follow-up that extracts a shared `buildAgentCommand()` would make this class of bug structurally harder to introduce.

### 3. Does this fix leave technical or structural debt?

**Some, which is normal for a targeted bug fix:**

- **Command-building logic** is still spread across 8 files — a natural candidate for extraction when someone's already working in that area.

- **`spawnWithTimeout()`** appears in 4 places with the same implementation — same consolidation opportunity.

- **Some docs and comments still reference `gh copilot`** while the code now calls `copilot` directly. Not urgent, but worth a docs-cleanup pass at some point.

None of these are blockers for merging — the PR does what it sets out to do. They're just opportunities for the next pass.

PR #1180, opened 2026-05-27
feat: Coordinator-as-Agent export — compile .squad/ into .github/agents/squad.md

bradygaster:
## Summary

Extends the `squad export` command family with a new **coordinator agent export** that compiles `.squad/` state into a repository-native Copilot custom agent at `.github/agents/squad.md`.

This lets Squad teams export their roster, routing rules, ceremony triggers, and dispatch protocol into a format that works across all Copilot surfaces (CLI, VS Code, GitHub Desktop) without requiring runtime changes.

### CLI

```bash
# New coordinator export
squad export agent
squad export agent --watch    # live re-export on .squad/ changes
squad export agent --check    # CI-friendly drift detection
squad export agent --dry-run  # preview without writing
squad export agent --compact  # force smaller output

# Existing snapshot export (unchanged)
squad export
squad export snapshot
```

### Architecture

The export pipeline:
1. **Context loader** — reads `.squad/team.md`, `routing.md`, `ceremonies.md`, `config.json`, and agent charters into a typed IR
2. **Prompt compiler** — renders a coordinator prompt with token budget management (soft: 14k, hard: 20k)
3. **Frontmatter renderer** — emits valid custom-agent YAML frontmatter
4. **File writer** — writes `.github/agents/squad.md` with safety checks (collision detection, generated-file markers, legacy cleanup)

### Token budget & compaction

- Normal teams (≤8 members): full prompt mode
- When over soft budget: automatic compact mode
- Large teams (>8 members or >3k roster tokens): lazy-load mode (instructs coordinator to read charters at dispatch time)
- Hard budget exceeded: fails with actionable diagnostics

### Safety

- Won't overwrite user-owned agent files unless `--force`
- Detects legacy `squad.agent.md` collisions
- Generated files are marked with a comment header for safe re-export
- `--check` mode enables CI enforcement without mutation

### New
- `packages/squad-sdk/src/repo-native/` — types, IR loader, prompt compiler, frontmatter renderer, file writer, watch mode
- `packages/squad-cli/src/cli/commands/export-coordinator.ts` — CLI entry point for agent export
- `test/repo-native/coordinator-export.test.ts` — 15 tests covering the full pipeline

### Modified
- `packages/squad-cli/src/cli/commands/export.ts` — forked into a snapshot/agent router
- `packages/squad-cli/src/cli-entry.ts` — updated help text and arg passing
- `packages/squad-sdk/src/index.ts` — re-exports repo-native module
[...]

PR #1187, opened 2026-05-28
feat(export): v008 quality-first viability gate

bradygaster:
## Summary

Implements the v008 quality-first philosophy for coordinator-as-agent export: **we'd rather refuse with honest advice than produce a compressed artifact that lacks enough context to be useful.**

### What's New

- **Viability gate** — Pre-flight check assessing squad complexity before attempting export
  - Agent count thresholds (>25 warn, >50 error)
  - Routing rule density thresholds (>60 warn, >100 error)  
  - Coverage analysis (structural content vs budget ratio)
  - All thresholds scale with configured character limit
- **Configurable output limit** — `--char-limit <n>` flag (default 30000)
  - Future-proofs for CCA limit increases
  - Safety margin of 1000 chars below the limit
- **`--force` escape hatch** — Override viability refusal with honest disclaimer
- **Roster parsing fix** — Filters out category headers, placeholder rows, and non-member entries from team.md tables
- **Dispatch rules distillation** — Groups routes to same target when >20 rules, caps principles at 5
- **Decision distillation** — Progressive compaction from full → compact → lazy-load → minimize → drop
- **Self-containment enforcement** — Strips all external file references CCA cannot access

### Viability Gate UX

When a squad is too complex:
```
We're pretty sure this squad is too complex to produce a valuable export within the 30000-char limit.

• This squad has 55 specialists — even in compact mode, the roster alone may consume most of the 30000-char budget.
• 120 routing rules is very dense. Even distilled, this many rules may produce a dispatch section that dominates the budget.

Use --force if you want to attempt it anyway — we'll do our best, but the output may lack enough context to be useful for an AI agent.
```

### Test Results

- 10 unit tests passing (viability check + roster parsing)
- Local squad repo: 24 members, 39 rules → viable, 5K/30K chars
- Simulated 55-agent ModSquad → correctly rejected (complexity: 97/100)
- `--force` correctly overrides rejection
- Configurable limit correctly scales thresholds (60K limit → 55 agents passes)

### CLI Usage

```bash
squad export agent                    # Standard export
squad export agent --char-limit 50000 # Custom limit
squad export agent --force            # Override viability refusal
squad export agent --compact          # Force compact mode
squad export agent --dry-run          # Preview without writing
squad export agent --check            # CI drift detection
```


bradygaster:
Consolidating into #1180 — all v008-v011 work lands there.

PR #1192, opened 2026-05-29
fix(permissions): use 'approve-once' for Copilot CLI v1.0.54+ contract

### LINKED ISSUE #1191: [bug] squad-cli cannot access tools due to Copilot CLI (post-v1.0.54) permission contract change
## What happened

Squad's CLI (and SDK) are no longer able to grant tool access when the environment has Copilot CLI v1.0.54 or newer. Any permission requests handled by the default handler (for tool use, shell, etc.) are not approved, blocking core functionality.

### Why
- The contract changed the expected `onPermissionRequest` handler return values as of Copilot CLI 1.0.54 and above.
- Instead of `{ kind: 'approved' }`, the SDK now expects `{ kind: 'approve-once' }` or another allowed value (see: references below)
- The current `approveAllPermissions` handler in `packages/squad-cli/src/cli/shell/index.ts` and other code paths still return `{ kind: 'approved' }`.
- Type definitions (e.g., `SquadPermissionRequestResult.kind`) and documented guidance also reference `approved` and other outdated strings.
- The Copilot CLI (post-v1.0.54) ignores these permission responses, so the squad CLI can't approve any tools.

## References:
- [Copilot SDK permission handler docs](https://github.com/github/copilot-sdk/blob/main/nodejs/README.md#permission-handling)
- [Code search: `approveAllPermissions` handler](https://github.com/bradygaster/squad/blob/dev/packages/squad-cli/src/cli/shell/index.ts)

bradygaster:
## Summary

Fixes #1191 — squad-cli cannot access tools due to Copilot CLI (post-v1.0.54) permission contract change.

## Problem

The Copilot CLI v1.0.54+ changed the `onPermissionRequest` handler contract. It now expects `{ kind: 'approve-once' }` instead of the deprecated `{ kind: 'approved' }`. The old value is silently ignored, blocking all tool access.

## Changes

- **`packages/squad-sdk/src/adapter/types.ts`** — Added `'approve-once'` to the `SquadPermissionRequestResult.kind` union type (kept `'approved'` for backward compat)
- **`packages/squad-cli/src/cli/shell/index.ts`** — Updated `approveAllPermissions` handler to return `'approve-once'`
- **`packages/squad-sdk/src/adapter/client.ts`** — Updated error message hint to reference `'approve-once'`

## Testing

- SDK type-checks clean (`tsc --noEmit` passes)
- CLI shell file type-checks clean against updated SDK dist
- No test files reference the old value

cc @tamirnavot @jonlester for review


jonlester:
@bradygaster 2 suggestions after reviewing:
- We should probably re-export the `approveAll` handler from copilot-sdk so that squad-sdk consumers (incl. squad-cli) can use that instead of hardcoding.
- To make this less brittle going forward, we should consider some version-pinned initialization validation.  Check `client.getStatus().protocolVersion` when  starting a session.  If the value is > 3 (the current version), log a warning to the console output.

tamirdresher:
Superseded by extended #1200 which now bundles commits 70a37812 (fix) + e0291f3f (test) plus WI-1, UPGRADE-FLAG-IGNORED, UPGRADE-NO-MIGRATION, and UPGRADE-EPERM-FALSE-SUCCESS. Recommend closing this after #1200 lands.

bradygaster:
@tamirdresher - did any of @jonlester's contributions fail to make it into that superceded pr? just making sure we got all my old boss' ideas. :)

tamirdresher:
@bradygaster — your old boss''s instincts were good, but neither suggestion made it into #1200.

What did land: your original `'approve-once'` fix from #1192, plus a backward-compat shim (commit `55e843c0`) normalizing legacy `'approved'` → `'approve-once'`. The kind-string area Jon flagged is covered.

The `approveAll` re-export didn''t land — `approveAllPermissions` is still a local hardcoded handler in `packages/squad-cli/src/cli/shell/index.ts` (~line 90). Every consumer still writes their own `() => ({ kind: 'approve-once' })` inline.

The `protocolVersion` check didn''t land either — `protocolVersion` exists as a type and flows through `getStatus()` and `ping()`, but no `if (… > 3) console.warn(…)` guard ever materialized.

Both are worth doing. I''ll open two follow-up issues and tag you and @jonlester — Jon can claim one if he wants to follow through on his own suggestions.

PR #1250, opened 2026-06-09
fix: release pipeline version pinning (#1203, #1204)

### LINKED ISSUE #1203: release-pipeline: squad-cli tarball declares unpublished SDK dependency version
**Discovered during:** local-tarball validation of PR #1200 (combined state-backend fixes).

**Symptom:** Installing the squad-cli tarball with `npm install -g <cli.tgz>` fails with:

```
npm error code ETARGET
npm error notarget No matching version found for @bradygaster/squad-sdk@>=0.9.6-preview
```

**Root cause:** The squad-cli tarball declares `""@bradygaster/squad-sdk"": "">=0.9.6-preview""` as a dependency, but no such version is published on npm. Per npm semver pre-release ordering, `insider` < `preview`, so the published `0.9.6-insider.3` does NOT satisfy the range. End users cannot install the tarball through any documented path.

**Real fixes (both desirable):**
1. Extend the release script to publish both `@bradygaster/squad-sdk` and `@bradygaster/squad-cli` atomically so the dependency range always resolves.
2. Optionally pin the SDK dependency in the CLI's `package.json` to an actually-published version when packing (or widen the range to `>=0.9.6-insider.3 || >=0.9.6-preview`).

**Workaround for validation only:** pack the SDK from the source workspace and install both side-by-side: `npm install -g <sdk.tgz> <cli.tgz>`. This is what iteration 3 of PR #1200 validation uses.

**Related:** #1098 (SDK pin during publish).

**Severity:** Blocks any external tarball-based smoke test of insider/preview cuts. Tagged `bug` because the released artifact is effectively uninstallable.

### LINKED ISSUE #1204: ensureSquadStateMcpPinned writes unresolvable version when CLI is a local-build / unpublished pre-release
## Root cause

`ensureSquadStateMcpPinned(dest, getPackageVersion())` (added in PR #1200, `packages/squad-cli/src/cli/core/upgrade.ts:705`) writes:

```json
{ "command": "npx", "args": ["-y", "@bradygaster/squad-cli@<cliVersion>", "state-mcp"] }
```

When `<cliVersion>` is **not on the npm registry** (every locally-built tarball, e.g. `0.9.6-preview.5`, and any developer's self-built CLI), npx fails at MCP-server startup:

```
npm error code ETARGET
npm error notarget No matching version found for @bradygaster/squad-cli@0.9.6-preview.5.
```

The MCP server never starts, the host registers zero tools, and agents correctly report `squad_state_*` unavailable. Orphan branch does not grow despite all the iter-3 fixes working at config-shape level.

## Evidence

- `npm view @bradygaster/squad-cli versions --json` highest published: `0.9.6-insider.3`. No `preview.5`.
- `npm view @bradygaster/squad-cli dist-tags --json`: `{ "latest": "0.9.4", "insider": "0.9.6-insider.3" }`
- Reproduced on `tamir-squad-hq-tarball-test-20260602T183202`. Full RCA report: `.squad/files/validation/MCP-LOADER-ROOT-CAUSE.md` in the validation companion repo.

## Recommended fix (Option A)

Validate the pin against the npm registry before writing. If unresolvable, fall back to `@bradygaster/squad-cli@insider` dist-tag. About 40 LOC in `upgrade.ts` plus a regression test in `test/mcp-bridge-pinning.test.ts` mocking both published-version and unpublished-version registry HEAD responses.

Affects PR #1200 / branch `squad/state-backend-upgrade-fixes` (this is the same change that introduced the function).

## Re-validation

After fix lands and a new twin tarball is built, recreate `tamir-squad-hq` dup, run `squad upgrade --state-backend two-layer`, run the 4 continuity prompts from `TARBALL-FULL-tamir-squad-hq.md` Phase 3.5. Expect `git log squad-state --oneline` to grow past the 2 baseline commits and Scribe to NOT print the `squad_state_*` unavailable halt.

## Why this is NOT a Copilot CLI bug

The MCP spec is invalid at npm-resolution time. No upstream change required.

bradygaster:
## Summary

Fixes two release pipeline version pinning bugs that surfaced in v0.10.0:

**Closes #1203** — CLI tarball declares unpublished SDK dependency version

The CLI's `package.json` declared `"@bradygaster/squad-sdk": ">=0.10.0"` — matching its own version as the floor. If the SDK publish fails or registry propagation is slow, `npm install` gets ETARGET because no version satisfies the range.

**Fix:** Lower the floor to `>=0.9.0` (the last known published version) so there's always a resolvable version. Also added a CI gate step that verifies the SDK dependency is actually resolvable on npm before publishing the CLI, preventing a repeat.

**Closes #1204** — `ensureSquadStateMcpPinned` writes unresolvable version for local/dev builds

The `buildMcpServerSpecs` function (used for agent frontmatter MCP config) directly wrote the CLI version into the spec without checking if it's published. Local dev builds (version `0.0.0`, build metadata like `0.10.0+local.1234`) would write unresolvable `npx` specs.

**Fix:** Added `isLocalOrUnpublishedVersion()` guard that detects placeholder versions (`0.0.0`), build metadata (`+`), and other local-build indicators — falling back to `@insider` dist-tag. Also extended `resolveSquadStateMcpSpec` to short-circuit for build-metadata versions.

## Changes

| File | Change |
|------|--------|
| `packages/squad-cli/package.json` | SDK dep floor: `>=0.10.0` → `>=0.9.0` |
| `packages/squad-cli/src/cli/core/upgrade.ts` | Added `isLocalOrUnpublishedVersion` guard to `buildMcpServerSpecs` |
| `packages/squad-cli/src/cli/core/mcp-spec.ts` | Short-circuit build-metadata versions |
| `.github/workflows/squad-npm-publish.yml` | Added SDK resolvability check before CLI publish |
| `test/mcp-spec-init.test.ts` | Tests for new guards |
| `.changeset/fix-release-pipeline-versions.md` | Changeset |

## Testing

- All 59 existing + new tests pass
- Build succeeds

### DIFF
--- a/.changeset/fix-release-pipeline-versions.md
+++ b/.changeset/fix-release-pipeline-versions.md
@@ -1,0 +1,14 @@
+---
+"@bradygaster/squad-cli": patch
+---
+
+fix: release pipeline version pinning (#1203, #1204)
+
+- Lower SDK dependency floor from `>=0.10.0` to `>=0.9.0` so the CLI tarball
+  resolves against the last published SDK when the current version isn't yet on
+  the registry.
+- Add `isLocalOrUnpublishedVersion` guard to `buildMcpServerSpecs` so local dev
+  builds and versions with build metadata (`+`) fall back to `@insider` instead
+  of writing an unresolvable version string into MCP config.
+- Extend `resolveSquadStateMcpSpec` to short-circuit for build-metadata versions.
+- Add CI step to verify SDK dependency is resolvable before publishing the CLI.
--- a/.github/workflows/squad-npm-publish.yml
+++ b/.github/workflows/squad-npm-publish.yml
@@ -316,6 +316,31 @@ jobs:
             exit 1
           fi
           echo "SDK dependency is safe: $SDK_DEP"
+      - name: "Verify SDK version is resolvable on npm (#1203)"
+        run: |
+          # Extract the minimum version floor from the SDK dependency range
+          SDK_DEP=$(node -p "require('./packages/squad-cli/package.json').dependencies['@bradygaster/squad-sdk']")
+          echo "SDK dependency declared: $SDK_DEP"
+          # Verify that at least one version satisfying the range exists on the registry
+          if ! npm view "@bradygaster/squad-sdk" versions --json 2>/dev/null | node -e "
+            const semver = require('semver');
+            const range = process.argv[1];
+            let data = '';
+            process.stdin.on('data', c => data += c);
+            process.stdin.on('end', () => {
+              const versions = JSON.parse(data);
+              const arr = Array.isArray(versions) ? versions : [versions];
+              const match = arr.find(v => semver.satisfies(v, range));
+              if (!match) {
+                console.error('::error::No published SDK version satisfies ' + range);
+                process.exit(1);
+              }
+              console.log('✅ SDK dependency resolves to: ' + match);
+            });
+          " "$SDK_DEP"; then
+            echo "::error::Cannot verify SDK dependency is resolvable — aborting CLI publish to prevent #1203"
+            exit 1
+          fi
       - name: Build squad-sdk (required for CLI project references)
         run: npm -w packages/squad-sdk run build
       - name: Build squad-cli
--- a/packages/squad-cli/package.json
+++ b/packages/squad-cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bradygaster/squad-cli",
-  "version": "0.10.0-build.2",
+  "version": "0.10.0",
   "description": "Squad CLI — Command-line interface for the Squad multi-agent runtime",
   "type": "module",
   "bin": {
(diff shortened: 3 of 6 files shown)


tamirdresher (approved):
✅ **Approving — release-engineering hygiene, tests pass cleanly.**

Built this branch locally. All 20 tests pass (`mcp-spec-init.test.ts` 14 + `mcp-root-write.test.ts` 6, including the new `isLocalOrUnpublishedVersion guard (#1204)` cases).

**Review:**

- `package.json` SDK dep floor `>=0.10.0` → `>=0.9.0` — the right call. Prevents the ETARGET race when CLI tarball ships before SDK propagates.
- `isLocalOrUnpublishedVersion()` correctly catches: empty, `0.0.0`, and SemVer build metadata (`0.10.0+local.1234`). Tests cover all three paths.
- `resolveSquadStateMcpSpec` short-circuits for build-metadata versions → falls back to `@insider`. Clean.
- CI workflow gate (`squad-npm-publish.yml`) — `npm view @bradygaster/squad-sdk versions` check before CLI publish is the right place to break the publish if propagation hasn't completed.

No concerns. Approving.

PR #1262, opened 2026-06-10
feat(init): prompt to add @copilot as a team member during squad init

### LINKED ISSUE #1147: UX: squad init should prompt to add @copilot as a team member
## Problem

When a user runs `squad init` and then invokes Copilot-based features (e.g. `copilot --agent squad -p ...`), they hit a silent gap: `.github/copilot-instructions.md` was never created, so Copilot has no Squad context and behaves like a plain assistant.

The current flow requires a separate explicit step:

```
squad init
squad copilot enable   # required — but easy to miss
copilot --agent squad -p 'build something'
```

This separation is **intentional by design**: `@copilot` is an opt-in team member and `.github/copilot-instructions.md` affects all Copilot interactions in the repo, so it should not be silently dropped during init.

However, the UX gap is real: new users who don't know about `squad copilot enable` will get a confusing first-run experience.

## Proposed fix

During `squad init`'s interactive setup flow (after team scaffolding completes), add a prompt:

```
Want to add @copilot as an autonomous team member? (yes/no)
```

- **yes** → run the equivalent of `squad copilot enable` inline
- **no / non-interactive** → skip silently; note in output that it can be added later with `squad copilot enable`

This would mirror the existing post-setup input-source prompts (PRD, GitHub issues, human members) already described in the coordinator prompt.

## Context

Discovered during E2E validation of PR #1035. The E2E test correctly ran `squad copilot enable` as a prerequisite for Copilot-integration scenarios — this is not a bug in the test or in `squad init`. But the note in the E2E verdict comment prompted this question, which surfaced the UX gap.

_filed with ghcp_

bradygaster:
## Summary

During interactive `squad init`, users are now prompted to add `@copilot` as an autonomous team member. This closes the UX gap where users had to know about `squad copilot enable` as a separate step.

## Behavior

- **Interactive (TTY):** After scaffolding completes, prompts: `Add @copilot as an autonomous team member? [Y/n]:`
  - **Yes (default):** Adds Coding Agent section to `team.md` and copies `copilot-instructions.md` template to `.github/`
  - **No:** Skips with a hint message about `squad copilot enable`
- **Non-interactive:** Skips silently with guidance message
- **Existing @copilot:** Skips entirely (idempotent)

## Testing

- Build passes (`npm run build`)
- Follows existing prompt patterns (same readline/promises approach as worktree strategy prompt)
- Uses same `insertCopilotSection`/`writeTeamMd` helpers as `squad copilot enable`

Closes #1147

### DIFF
--- a/.changeset/init-prompt-copilot-member.md
+++ b/.changeset/init-prompt-copilot-member.md
@@ -1,0 +1,10 @@
+---
+"@bradygaster/squad-cli": minor
+---
+
+Add @copilot team member prompt during `squad init`
+
+During interactive `squad init`, users are now prompted to add @copilot
+as an autonomous team member. Answering yes adds the Coding Agent section
+to team.md and copies copilot-instructions.md into the project. Non-interactive
+mode skips silently with a hint to run `squad copilot enable` later.
--- a/packages/squad-cli/src/cli/core/init.ts
+++ b/packages/squad-cli/src/cli/core/init.ts
@@ -17,6 +17,12 @@ import { liftInitMutableStateOntoOrphan } from '../commands/migrate-backend.js';
 import { resolveSquadStateMcpSpec } from './mcp-spec.js';
 import { describeMcpSpec } from './upgrade.js';
 import { ensureSquadStateMcpInRoot, ensureSquadStateMcpInUserConfig, tombstoneStaleSquadStateInProjectMcp } from './mcp-root.js';
+import {
+  readTeamMd,
+  writeTeamMd,
+  hasCopilot,
+  insertCopilotSection,
+} from './team-md.js';
 
 const storage = new FSStorageProvider();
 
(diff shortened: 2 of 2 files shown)


tamirdresher (approved):
Reviewed the 48-line addition to `init.ts`.

- TTY check via `process.stdin.isTTY` — correct, handles non-interactive gracefully
- Reuses existing `hasCopilot`, `insertCopilotSection`, `writeTeamMd` helpers — no duplicate logic
- Idempotent: `hasCopilot()` short-circuits if @copilot already on the roster
- Default to Yes — good UX for new users
- Copies `copilot-instructions.md` from templates, gated on `!storage.existsSync(instructionsDest)` so it never overwrites existing custom instructions
- Non-interactive path prints discoverable hint about `squad copilot enable`

Minor non-blocking observation: the `path.resolve(currentFilePath, '..', '..', '..', 'templates')` pattern depends on dist file location. Existing pattern in this file, so OK to follow. Worth a utility helper at some point.

CI green. Approving.

PR #1263, opened 2026-06-10
chore: simplify workflow files by removing .ai-team fallback logic

### LINKED ISSUE #1167: enhancement: Simplify workflow files
When setting up squad on a repository, it creates a lot of workflow files.
[see this PR for a recent clean init outcome](https://github.com/kubefleet-dev/kubefleet/pull/724/changes#diff-1a7f3f255414f1f033abb0d0b54f3a95f6f9b03e4b34ca3f31de5dbdaa840ff0)
these files embedd a fair bit of logic.
It would be a cleaner if the squad cli owned the logic, and the workflow files just invoked it using something like

`squad workflow <workflow-name>`

This would also allow to package and version the workflow logic itself together with the squad cli release, and remove some template upgrade requirements.

For example, logic like below does not belong to the workflow artifact IMO:

```
   // Read team roster — check .squad/ first, fall back to .ai-team/
            let teamFile = '.squad/team.md';
            if (!fs.existsSync(teamFile)) {
              teamFile = '.ai-team/team.md';
            }
            if (!fs.existsSync(teamFile)) {
              core.warning('No .squad/team.md or .ai-team/team.md found — cannot assign work');
              return;
            }

```

or: 

```
if [ -f ".squad/templates/ralph-triage.js" ]; then
            echo "has_script=true" >> $GITHUB_OUTPUT
          else
            echo "has_script=false" >> $GITHUB_OUTPUT
            echo "⚠️ ralph-triage.js not found — run 'squad upgrade' to install"
          fi
```

Driving the entire workflow logic through the squad CLI also guarantees that you can run it locally and test it easily.
if a user wants a custom workflow, they will still be able to create a fully customized one by scripting it.

bradygaster:
## Summary

Removes dead `.ai-team/` fallback code from workflow templates. The `.squad/` directory has been the canonical location since the rename — these fallback paths were never triggered and added unnecessary complexity to every workflow file.

## Changes

- Removed `.ai-team/` fallback branches from 4 workflow templates (triage, issue-assign, heartbeat, sync-squad-labels)
- Cleaned up empty if-blocks left by the removal
- Simplified warning messages to reference only `.squad/team.md`
- Changed `let` → `const` for file paths that are no longer reassigned
- Synced all template copies (`.squad-templates/`, `packages/squad-cli/templates/`, `packages/squad-sdk/templates/`)

## Not Changed

Guard rail checks in `squad-preview.yml` and `squad-promote.yml` are intentionally preserved — they legitimately prevent legacy `.ai-team/` files from being shipped to production.

## Scope Note

The broader simplification (moving workflow logic into CLI commands like `squad workflow <name>`) is a larger architectural change tracked separately. This PR focuses on the quick win: removing dead code paths that add complexity without value.

Closes #1167

### DIFF
--- a/.squad-templates/workflows/squad-heartbeat.yml
+++ b/.squad-templates/workflows/squad-heartbeat.yml
@@ -106,10 +106,7 @@ jobs:
           script: |
             const fs = require('fs');
 
-            let teamFile = '.squad/team.md';
-            if (!fs.existsSync(teamFile)) {
-              teamFile = '.ai-team/team.md';
-            }
+            const teamFile = '.squad/team.md';
             if (!fs.existsSync(teamFile)) return;
 
             const content = fs.readFileSync(teamFile, 'utf8');
@@ -154,7 +151,7 @@ jobs:
                     agent_assignment: {
                       target_repo: `${context.repo.owner}/${context.repo.repo}`,
                       base_branch: repoData.default_branch,
-                      custom_instructions: `Read .squad/team.md (or .ai-team/team.md) for team context and .squad/routing.md (or .ai-team/routing.md) for routing rules.`
+                      custom_instructions: `Read .squad/team.md for team context and .squad/routing.md for routing rules.`
                     }
                   });
                   core.info(`Assigned copilot-swe-agent[bot] to #${issue.number}`);
--- a/.squad-templates/workflows/squad-issue-assign.yml
+++ b/.squad-templates/workflows/squad-issue-assign.yml
@@ -27,13 +27,9 @@ jobs:
             // Extract member name from label (e.g., "squad:ripley" → "ripley")
             const memberName = label.replace('squad:', '').toLowerCase();
 
-            // Read team roster — check .squad/ first, fall back to .ai-team/
-            let teamFile = '.squad/team.md';
+            const teamFile = '.squad/team.md';
             if (!fs.existsSync(teamFile)) {
-              teamFile = '.ai-team/team.md';
-            }
-            if (!fs.existsSync(teamFile)) {
-              core.warning('No .squad/team.md or .ai-team/team.md found — cannot assign work');
+              core.warning('No .squad/team.md found — cannot assign work');
               return;
             }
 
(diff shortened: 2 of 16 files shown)


tamirdresher (approved):
Read the 16-file diff. Pure dead-code removal — strips `.ai-team/` fallback branches from 4 workflow templates (heartbeat, issue-assign, triage, sync-squad-labels), keeps the `.squad/` paths. Synced across all 4 template directories. Correct.

Confirmed the guard-rail keep-out reasoning: `squad-preview.yml` and `squad-promote.yml` are intentionally LEFT alone because they specifically prevent legacy `.ai-team/` files from shipping. Good call.

**Policy Gates CI failure is a real but trivial gap:**

```
::error::No changeset or CHANGELOG.md update found, but SDK/CLI source or template files were changed.
::error::Run 'npx changeset add' or edit CHANGELOG.md. Escape hatch: add 'skip-changelog' label.
```

The PR touches template files which fall under the changelog-gate's scope. Options:

- **(recommended)** Add `skip-changelog` label — this is purely dead-code removal with no user-visible behavior change, so a changeset entry would just be noise
- Alternatively, add a tiny changeset patch entry

Approving the change; merge gate just needs the label or a changeset add.

tamirdresher:
**Status:** Content approved. Only blocker is the **Changelog Gate** in Policy Gates:

```
::error::No changeset or CHANGELOG.md update found, but SDK/CLI source or template files were changed.
::error::Run 'npx changeset add' or edit CHANGELOG.md. Escape hatch: add 'skip-changelog' label.
```

This PR is purely dead-code removal (the `.ai-team/` fallback paths in workflow templates) with **no user-visible behavior change** — a changeset entry would be noise.

### Recommended unblock

Add the `skip-changelog` label. The gate explicitly documents this as the right escape hatch for non-user-facing changes.

Alternative: add a one-line `.changeset/` entry, but it would just say "internal cleanup, no user-visible change" which is what the label communicates.

### Other notes

The PR correctly leaves `squad-preview.yml` and `squad-promote.yml` untouched — those guard rails actively prevent legacy `.ai-team/` files from shipping to production. Good call.

Once the label is on, this is mergeable.

PR #1295, opened 2026-06-13
feat(sdk): bundle cross-squad-communication skill (#5 cross-squad protocols)

tamirdresher:
PR #1291 added `squad registry add/list/remove` and the `cross-squad/SKILL.md` for peer discovery. But discovery is only half the story — once a peer is known, agents need to know **how** to actually exchange info with it.

This change ports `cross-squad-communication` from [tamirdresher/squad-skills](https://github.com/tamirdresher/squad-skills/tree/main/plugins/cross-squad-communication) into the bundled skills so a fresh `squad init` produces a coordinator that knows the four communication patterns.

## What this skill teaches

| Pattern | When to use |
|---|---|
| **Pattern 0**: Synchronous CLI session | Quick knowledge queries — spawn `copilot` with `--working-directory` set to target repo |
| **Pattern 1**: Read-only metadata scan | "What's the architecture of squad X?" — read their `team.md` / `decisions.md` directly |
| **Pattern 2**: Async git-based request/response | Long-running work, PR reviews, multi-cycle tasks. Files in `.squad/cross-squad/requests/` + `.squad/cross-squad/responses/` |
| **Pattern 3**: Issue-based delegation | GitHub-hosted repos — `gh issue create` with `squad:cross-squad` label as the message bus |

Plus: decision tree for choosing the right pattern, anti-patterns, request/response YAML format, validation status.

## Changes

- **New** `.squad/skills/cross-squad-communication/SKILL.md` (379 lines, canonical source). `sync-skill-templates.mjs` (which runs in `prebuild`) propagates to both `packages/squad-cli/templates/skills/` and `packages/squad-sdk/templates/skills/`.
- **`MANIFEST_SKILL_NAMES`** in `packages/squad-sdk/src/config/init.ts` grows by 1 entry: `cross-squad-communication` → 11 entries total.
- **`cross-squad/SKILL.md`** (the registry-aware skill from #1291) gets a one-paragraph "Companion skill" note at the top pointing to `cross-squad-communication` for protocol details. The two are designed to be used together: `cross-squad` answers **"who?"** (discovery via registry), `cross-squad-communication` answers **"how?"** (the 4 communication patterns).

## Genericization

The original plugin documented validation against specific internal Microsoft repositories. Examples in this version use generic names (`platform-squad`, `research-squad`, etc.) so they are meaningful to all upstream users. The protocol mechanics are unchanged. The frontmatter `source:` attributes the port to `tamirdresher/squad-skills`.

## Test coverage

[...]

PR #1292, opened 2026-06-13
fix(sdk,cli): bundle missing skills on init + strip fabricated tiered-memory provenance (#1289, #1264)

tamirdresher:
Two regressions shipped in v0.10.0 caused several skills to silently never reach users. This PR fixes them in one place and adds a CI-style guard so they cannot recur.

## What's broken on `dev` today

| # | Bug | Mechanism |
|---|-----|-----------|
| [#1289](https://github.com/bradygaster/squad/issues/1289) | `squad-commands` + `squad-version-check` never install (the `squad commands` trigger phrase is a dead no-op on every v0.10.0 install) | Both are listed in `MANIFEST_SKILL_NAMES` (`packages/squad-sdk/src/config/init.ts`) but the source dirs are only in `packages/squad-cli/templates/skills/`, not `packages/squad-sdk/templates/skills/`. The install loop at `sdk-init.ts:1176` does `if (storage.existsSync(srcSkill))` and silently skips missing dirs. |
| [#1264](https://github.com/bradygaster/squad/issues/1264) | `tiered-memory`, `iterative-retrieval`, `reflect` never install | Source files exist in both templates dirs but the skill names were never added to `MANIFEST_SKILL_NAMES` |
| [#1264](https://github.com/bradygaster/squad/issues/1264) | `tiered-memory/SKILL.md` claims fabricated production measurements | Frontmatter says `confidence: high` and `source: earned (production measurements in tamirdresher/tamresearch1, 34-74KB baseline payloads)`. That repo does not exist. The Measurement Data table at the bottom of the file is fabricated. Also references non-existent `docs/tiered-memory-guide.md`. |
| (no issue filed) | `cross-squad` skill (updated in #1291) never installs | #1291 updated the SKILL.md but didn't add `cross-squad` to `MANIFEST_SKILL_NAMES` |

## Fix

1. **Canonical source corrected.** Adds the 5 missing skill directories under `.squad/skills/` (`squad-commands`, `squad-version-check`, `tiered-memory`, `iterative-retrieval`, `reflect`). The pre-existing `scripts/sync-skill-templates.mjs` (which runs in `prebuild`) propagates them to both packages' templates dirs.
2. **Manifest grown** 10 → 14 entries: adds `tiered-memory`, `iterative-retrieval`, `reflect`, `cross-squad`.
3. **Anti-regression guard.** The install loop now collects missing source dirs and throws with a clear remediation message (`Run \
[...]

PR #1297, opened 2026-06-13
fix(skills): add 'squad' disambiguation skill so skill(Squad) lookups succeed

tamirdresher:
## Symptom

Observed in a fresh `squad init` session, 2026-06-13:

```
◐ The Squad skill is perfect for assembling a specialized team to build C# console applications, so I'll use that to get the right specialists set up in the repo.
✗ skill(Squad)   Skill not found: Squad
◐ I see the issue—Squad is actually a custom agent type in the task tool, not a standalone skill, so I need to adjust my approach and call the task tool with agent_type set to "Squad" instead.
```

The coding model burns a turn on a failed skill lookup before self-correcting. The pattern is reproducible across models because the Squad agent's description (*"Your AI team. Describe what you're building, get a team of specialists that live in your repo."*) reads exactly like a skill description.

## Root cause

- `.github/agents/squad.agent.md` declares `name: Squad` — that is the **custom agent**.
- All bundled skills are hyphenated lowercase: `squad-commands`, `squad-conventions`, `squad-version-check`, etc.
- Nothing matches the literal name `Squad` for skill lookup, so the model's first guess fails.

## Fix

Add a small `squad` skill that exists purely to **catch the bad lookup and redirect**. It does no work — it explicitly tells the model:

- Squad is a **custom agent**, not a skill.
- To invoke the coordinator: `task(name=…, agent_type="Squad", prompt=…)`
- To list commands: trigger the `squad-commands` skill (`squad commands` / `what can squad do` / etc.)
- To init in a new project: `squad init` in the shell (not from inside a session)

Triggers cover the natural-language phrasings that would lead the model to mis-call `skill(Squad)` (`squad`, `use squad`, `use the squad skill`, `squad skill`, `assemble a squad`, `set up squad`, `squad agent`, `squad team`).

## Changes

- New `.squad/skills/squad/SKILL.md` (canonical source). `sync-skill-templates.mjs` (prebuild) propagates to both packages templates.
- `MANIFEST_SKILL_NAMES` in `packages/squad-sdk/src/config/init.ts` adds one entry: `'squad'` → 11 entries total.

## Test coverage

New `test/init.test.ts > should install the squad disambiguation skill` asserts:
- File exists at `.copilot/skills/squad/SKILL.md` after `initSquad()`
- Content contains the literal string `CUSTOM AGENT` (warns the model)
- Content contains the correct invocation pattern `agent_type="Squad"`
- Content references `squad-commands` (the right skill for the menu use case)

26/26 init tests pass; `npm run lint` clean.

## Composability with other open PRs

Disjoint from #1292 (adds 4 entries) and #1295 (adds 1 entry). All three can land in any order. Merge order doesn't matter.

## Verified end-to-end

Built and tested locally — Copilot session in a fresh init now finds `squad` skill on first attempt, reads the redirect, then issues the correct `task(agent_type="Squad", …)` call instead of failing.

### DIFF
--- a/.changeset/add-squad-disambiguation-skill.md
+++ b/.changeset/add-squad-disambiguation-skill.md
@@ -1,0 +1,55 @@
+---
+"@bradygaster/squad-sdk": patch
+"@bradygaster/squad-cli": patch
+---
+
+Fix: add `squad` disambiguation skill so `skill(Squad)` lookups succeed and redirect to the right tool
+
+**Symptom (observed in a fresh `squad init` session, 2026-06-13):**
+
+```
+◐ The Squad skill is perfect for assembling a specialized team to build C# console applications, so I'll use that to get the right specialists set up in the repo.
+✗ skill(Squad)   Skill not found: Squad
+◐ I see the issue—Squad is actually a custom agent type in the task tool, not a standalone skill, so I need to adjust my approach and call the task tool with agent_type set to "Squad" instead.
+```
+
+The model burns a turn on a failed skill lookup before self-correcting. The pattern is reproducible across coding models because the Squad agent's description (`"Your AI team. Describe what you're building, get a team of specialists that live in your repo."`) reads exactly like a skill description.
+
+**Root cause:**
+
+- `.github/agents/squad.agent.md` declares `name: Squad` — that's the **custom agent**.
+- All bundled skills are hyphenated lowercase: `squad-commands`, `squad-conventions`, `squad-version-check`, etc.
+- Nothing matches the literal name `Squad` for skill lookup.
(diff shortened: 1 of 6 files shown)


tamirdresher:
Superseded by #1302 — the squad skill name in #1297 collided with the Copilot agent named Squad and never appeared in /skills. After decompiling Copilot CLI 1.0.62-2 app.js, confirmed the skill loader's name field would have collided with the agent name. PR #1302 renames to squad-help to avoid the collision, removes the unused 	riggers: frontmatter (Copilot CLI ignores it), and also adds the missing TEMPLATE_MANIFEST entry so squad upgrade propagates the skill (a separate code path from squad init).

PR #1298, opened 2026-06-13
fix(cli): stop writing squad_state to ~/.copilot/mcp-config.json on every init/upgrade (#1296)

### LINKED ISSUE #1296: squad init writes squad_state_<hash> to ~/.copilot/mcp-config.json on every init, contradicting iter-8 design
`squad init` (at any state backend) unconditionally writes `squad_state_<hash>` to `~/.copilot/mcp-config.json` via `ensureSquadStateMcpInUserConfig` (`packages/squad-cli/src/cli/core/init.ts:408`). One entry per project accumulates in HOME with no GC.

This contradicts the explicit iter-8 design intent documented at `packages/squad-cli/src/cli/core/mcp-root.ts:1-27`:

> *iter-7: wrote `squad_state_<hash>` into the user's HOME `~/.copilot/mcp-config.json`. That polluted HOME with one entry per Squad project and required a stale-entry GC that we never built. It also touched a file outside the project, which is surprising for `squad init` / `squad upgrade`.*
>
> *iter-8 flips it back inside the project: we write `squad_state` to a repo-root `.mcp.json` ... **No HOME modifications.***

But `init.ts:408` then turns around and modifies HOME anyway. The accompanying comment `// Also pin to user-level config for external 'copilot -p' compatibility` claims this is for `copilot -p` mode, but `copilot -p` reads `.mcp.json` from the cwd too — the repo-root write at line 403 is already sufficient for that case.

**Repro**

`powershell
mkdir C:\temp\squad-home-pollution-repro; cd C:\temp\squad-home-pollution-repro; git init
squad init
(Get-Content ~/.copilot/mcp-config.json -Raw | ConvertFrom-Json).mcpServers.PSObject.Properties.Name | Where-Object {  -match '^squad_state_' }
# → squad_state_<hash>   ← polluted
`

**Impact**

- Each new `squad init` adds another `squad_state_<hash>` entry; stale entries point at deleted/moved projects
- The keys are opaque hashes, so users have no obvious way to clean them
- `squad upgrade` (`upgrade.ts:738`) ALSO calls `ensureSquadStateMcpInUserConfig`, amplifying the problem
- Defeats the iter-8 goal of "init/upgrade should not modify HOME"

**Fix**

Remove the two `ensureSquadStateMcpInUserConfig` calls (init.ts:408 and upgrade.ts:738). The repo-root `.mcp.json` writes (line 403 / line 728) cover all documented Copilot CLI launch modes including `copilot -p` from the project root.

For users who do invoke `copilot -p` from somewhere other than the project root, the right fix is to teach them `--additional-mcp-config @.mcp.json` (which `init.ts:494` already documents as the recommended pattern).

**Cleanup helper**

A follow-up could add `squad doctor --mcp-prune` to walk `~/.copilot/mcp-config.json`, find `squad_state_<hash>` entries whose target directory no longer exists, and remove them.

cc: discovered in the v0.10 stabilization weekend session.

tamirdresher:
Fixes #1296.

## Symptom

`squad init` (line 408 of `init.ts`) and `squad upgrade` (line 738 of `upgrade.ts`) unconditionally call `ensureSquadStateMcpInUserConfig`, writing `squad_state_<hash>` to `~/.copilot/mcp-config.json` keyed by a stable hash of the project path. Every new `squad init` accumulates another entry in HOME with no garbage collection.

Verified live on 2026-06-13: a user's HOME mcp-config had two stale `squad_state_*` entries from two different project paths (only one of which still exists on disk).

## Why this is wrong

Contradicts the explicit iter-8 design intent at `packages/squad-cli/src/cli/core/mcp-root.ts:1-27`:

> *iter-7: wrote `squad_state_<hash>` into the user's HOME `~/.copilot/mcp-config.json`. That polluted HOME with one entry per Squad project and required a stale-entry GC that we never built. It also touched a file outside the project, which is surprising for `squad init` / `squad upgrade`.*
>
> *iter-8 flips it back inside the project: we write `squad_state` to a repo-root `.mcp.json` ... **No HOME modifications.***

The repo-root writes (init.ts:403 / upgrade.ts:728) already cover all documented Copilot CLI launch modes. Out-of-tree `copilot -p` invocations should use `--additional-mcp-config @.mcp.json` (already documented at init.ts:494 as the recommended pattern).

## Changes

- Removed the unconditional `ensureSquadStateMcpInUserConfig` call from `init.ts:408` and `upgrade.ts:738`. Replaced both with comments explaining the iter-8 design and pointing at #1296.
- Removed the now-unused import from both files.
- The function definition itself (`mcp-root.ts:178-228`) is **kept** — a future `squad doctor --mcp-prune` cleanup helper may want to inspect HOME for orphan entries. It's just no longer called from the init/upgrade auto-flow.

## Test coverage

New regression test in `test/cli/init.test.ts`:

> `should NOT write any squad_state entries to ~/.copilot/mcp-config.json (regression: #1296)`

Isolates the developer's real HOME by setting `USERPROFILE`/`HOME` to a temp dir before init, then asserts no `squad_state*` keys appear under that temp HOME after init.

All 40 existing init tests still pass; `npm run lint` clean.

## Cleanup follow-up (out of scope)

Add `squad doctor --mcp-prune` to walk `~/.copilot/mcp-config.json`, find `squad_state_<hash>` entries whose target directory no longer exists, and remove them.

### DIFF
--- a/.changeset/fix-1296-stop-mcp-home-pollution.md
+++ b/.changeset/fix-1296-stop-mcp-home-pollution.md
@@ -1,0 +1,40 @@
+---
+"@bradygaster/squad-cli": patch
+---
+
+Fix #1296: stop writing `squad_state_<hash>` to `~/.copilot/mcp-config.json` on every `squad init` / `squad upgrade`
+
+`squad init` (line 408 of `packages/squad-cli/src/cli/core/init.ts`) and `squad upgrade` (line 738 of `upgrade.ts`) unconditionally called `ensureSquadStateMcpInUserConfig`, which wrote a `squad_state_<hash>` entry to `~/.copilot/mcp-config.json` keyed by a stable hash of the project path. Every new `squad init` accumulated another entry in HOME with no garbage collection.
+
+This contradicted the explicit iter-8 design intent documented at `packages/squad-cli/src/cli/core/mcp-root.ts:1-27`:
+
+> *iter-7: wrote `squad_state_<hash>` into the user's HOME `~/.copilot/mcp-config.json`. That polluted HOME with one entry per Squad project and required a stale-entry GC that we never built. It also touched a file outside the project, which is surprising for `squad init` / `squad upgrade`.*
+>
+> *iter-8 flips it back inside the project: we write `squad_state` to a repo-root `.mcp.json` ... **No HOME modifications.***
+
+The repo-root `.mcp.json` writes (init.ts:403 / upgrade.ts:728) already cover all documented Copilot CLI launch modes:
+
+- ``copilot`` from the project root → walks up from cwd to git root, auto-loads `.mcp.json`
+- ``copilot -p`` from the project root → same `.mcp.json` is found
+
+For ``copilot -p`` invocations launched from **outside** the project root, the right pattern is `--additional-mcp-config @.mcp.json` (already documented as the recommended pattern at init.ts:494).
+
+**Changes**
+
+- Removed the unconditional `ensureSquadStateMcpInUserConfig` call from `init.ts:408` and `upgrade.ts:738`. Replaced both with comments explaining the iter-8 design and pointing at #1296.
+- Removed the now-unused import from both files.
+- The function definition itself (`mcp-root.ts:178-228`) is **kept** — a future `squad doctor --mcp-prune` cleanup helper may want to inspect HOME for orphan entries. It's just no longer called from the init/upgrade auto-flow.
+
+**Tests**
+
+New regression test in `test/cli/init.test.ts`:
+
+> `should NOT write any squad_state entries to ~/.copilot/mcp-config.json (regression: #1296)`
+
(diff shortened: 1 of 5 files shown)


tamirdresher:
Both reviewer threads addressed in commit 98893d07:

- **`Copilot CLI 5.3+` comments** — replaced all 5 occurrences (init.ts, upgrade.ts, mcp-root.ts) with `Copilot CLI ≥1.0.59` to match the real shipping version (current is 1.0.62).
- **Test comment with hardcoded line numbers** — rewrote the test's regression comment to reference the call by function name (`ensureSquadStateMcpInUserConfig`) instead of `init.ts:408` / `upgrade.ts:738`, so it stays accurate under unrelated edits.

Regression: #1296 test still passes locally.

bradygaster (approved):
✅ Flight approves. This restores the iter-8 no-HOME invariant, keeps the future prune path available, and adds a regression test for the exact pollution path.

PR #1306, opened 2026-06-13
fix(prompt): coordinator must probe for squad_state/memory tools before mutating state (closes #1305)

### LINKED ISSUE #1305: Coordinator does not probe for squad_state/memory tools — silently uses raw file ops on non-local state backends
## Symptom

A coordinator session was opened against a Squad with `stateBackend: "two-layer"`. The coordinator wrote `.squad/decisions.md`, four `.squad/agents/{name}/history.md` files, and a `.squad/decisions/inbox/copilot-directive-…md` file using raw `create`/`edit` tools — **never calling any `squad_state_*` or `memory.*` tool**. The pre-commit hook caught the contract violation; the agent treated it as a "git problem" instead of recognizing it as the symptom it was.

Self-analysis from the agent that did it (verbatim):

> *"The tools weren't visible in my default function list. My session loaded with `powershell`, `view`, `create`, `edit`, `grep`, `glob`, `task`, `store_memory`, `vote_memory`, etc. — the squad MCP tools (`squad_state_*`) were registered but not advertised in the tool definitions block at the top of my prompt. I only found them just now by running `tool_search_tool_regex`."*
>
> *"I interpreted 'not available' [in `squad.agent.md`] as 'not in my visible tool list' instead of proactively probing. The instant I saw `stateBackend: two-layer` in `config.json`, I should have stopped and confirmed the runtime tools existed."*

## Root cause (two failure modes stacked)

1. **Mechanical (Copilot CLI behavior):** Copilot CLI loads MCP server tools but doesn't preload them all into the model's initial function list — they're behind `tool_search_tool_regex`-style lazy discovery. Confirmed empirically: `.mcp.json` has `squad_state` registered with `tools: ["*"]`, server responds correctly to `initialize` and `tools/list` via JSON-RPC, but model's tool block doesn't list them upfront.

2. **Behavioral (squad.agent.md is too soft):** `squad.agent.md` says *"When memory tools are available, use them before writing durable memory by hand"* and *"If memory.write is not available, fall back to `squad_decide` or `squad_state_write`"*. The conditional `"available"` is ambiguous — models interpret it as "if listed in my tool prompt" instead of "after probing to find out". There's no hard refusal clause for the case where the agent is about to violate the state-backend contract.

## Why this matters for non-local backends

For `stateBackend: "two-layer"`, `orphan`, or `git-notes`, mutable state files are NOT on disk in the working tree — they live in the orphan branch / git notes / external store. Writing to `.squad/decisions.md` via `create`/`edit` either:
- Hits the pre-commit hook and fails the user
- Or worse, succeeds locally and creates phantom state that the runtime overwrites at next read

This is a **contract violation that the agent cannot recover from** without manual cleanup. The wrong-tool problem must be prevented at the prompt level, not detected at commit time.

## Proposed fix

Three layers (one is out of our control; two we can ship):

### 1. Session-init probe step in `squad.agent.md` (in scope)

After resolving STATE_BACKEND, the coordinator MUST probe for `squad_state_*` availability. Pseudo-procedure inserted near the top of the Coordinator Identity section:

> **State-backend handshake (run on every session, before any state mutation):**
>
> 1. Read `.squad/config.json`. If `stateBackend` is absent or `"local"`/`"worktree"`, skip the rest of this step — file ops on `.squad/` are valid for these backends.
> 2. If `stateBackend` is `"two-layer"`, `"orphan"`, or `"git-notes"`: probe for `squad_state_health` via `tool_search_tool_regex` (or any model-equivalent tool-discovery mechanism). If unavailable, **halt** — output: *"Squad's runtime state bridge is missing for backend `{name}`. The `squad_state` MCP server is not advertised in this session. Restart Copilot CLI with `.mcp.json` loaded, or set `stateBackend` to `local` in `.squad/config.json`."* — and stop until the user acknowledges.

### 2. Hard refusal clause in the State Backend section (in scope)

> **Hard rule:** If `stateBackend ∈ {two-layer, orphan, git-notes}` and the `squad_state_*` tools are not loaded, you MUST NOT write to `.squad/decisions.md`, `.squad/decisions/inbox/`, `.squad/agents/*/history.md`, `.squad/casting/*.json`, `.squad/identity/*.md`, or any `.squad/memory/*` path using `create`/`edit`/`write_file` tools. Report the missing bridge instead. This is a contract violation; the runtime owns persistence for these backends.

### 3. Auto-advertise MCP tools at handshake (OUT OF SCOPE for Squad — file as Copilot CLI feature request)

This is the only complete fix for the mechanical root cause, but it's a Copilot CLI architectural decision (lazy-discovery vs. preload). Out of Squad's control. Will file as a separate request against `github/copilot-cli` if maintainers agree it's worth it. With #1 + #2 in place, this becomes a nice-to-have rather than load-bearing.

## Also worth doing (small, in scope)

4. Reinforce the same rules in `.squad/skills/init-mode/SKILL.md` and `.squad/skills/agent-conduct/SKILL.md` so even a stale `squad.agent.md` install gets the right behavior via the skill loader.

## Acceptance criteria

- A fresh `squad init --state-backend two-layer` + `copilot --agent squad` session, opened with the upgraded coordinator, MUST attempt to probe for `squad_state_*` tools before writing any state file. If unavailable, MUST halt and tell the user.
- Test by stripping `squad_state` from `.mcp.json` and verifying the coordinator refuses to write `decisions.md` instead of trying.

tamirdresher:
---
"@bradygaster/squad-cli": minor
"@bradygaster/squad-sdk": minor
---

Fix #1305: coordinator must probe for `squad_state` / `memory.*` tools before writing state, and hard-refuse writes when the bridge isn't reachable on non-local backends

## Symptom

A coordinator session against a Squad with `stateBackend: "two-layer"` wrote `.squad/decisions.md` + four `.squad/agents/{name}/history.md` files + a `.squad/decisions/inbox/copilot-directive-…md` via raw `create`/`edit` tools — never calling any `squad_state_*` or `memory.*` tool. The pre-commit hook caught the contract violation; the agent treated it as a "git problem" instead of the symptom it was.

## Root cause (two failure modes stacked)

**Mechanical (Copilot CLI):** Copilot CLI loads MCP server tools lazily — they're registered (via `.mcp.json`) but not always advertised in the model's initial function list. Models have to use `tool_search_tool_regex` to find them or guess they exist. The `squad_state` MCP server is correctly registered with `tools: ["*"]` and responds to `initialize`/`tools/list` over stdio — but the model's tool block at session start may not include them. This is a Copilot CLI architectural choice (lazy discovery vs. preload), not something Squad can fix server-side.

**Behavioral (squad.agent.md):** The pre-1305 prompt said *"When memory tools are available, use them before writing durable memory by hand"* and *"If memory tools are not available, fall back to squad_decide or squad_state_write"*. Models read `"available"` as `"listed in my tool prompt"` instead of `"after probing to find out"`. There was no hard refusal clause for the case where the agent is about to violate the state-backend contract.

## Fix

Two changes to `.squad-templates/squad.agent.md` (synced to all 4 mirror targets):

### 1. New "State-backend handshake" section (MANDATORY, runs once per session)

Inserted right after the `stateBackend` resolution at L131. Steps:

1. If `STATE_BACKEND ∈ {"local", "worktree"}`: file ops on `.squad/` are valid; skip the probe.
2. Otherwise: probe for `squad_state_health` via `tool_search_tool_regex` (or equivalent tool-discovery mechanism). On success, call `squad_state_health` once to confirm the bridge answers.
3. **If the probe fails**: HALT before any state write. Output a precise error to the user (verbatim text in the template): *"Squad's runtime state bridge is missing for backend `{STATE_BACKEND}`. The `squad_state` MCP server in `.mcp.json` is not reachable in this Copilot session. Restart Copilot CLI so `.mcp.json` is loaded, or change `stateBackend` to `local` in `.squad/config.json`."*

### 2. Replaced soft "if not available" language with a HARD RULE in `### Memory Governance Tools`

Lists the runtime-owned paths that are FORBIDDEN to write via `create`/`edit`/`write_file` on non-local backends when the bridge isn't reachable:

- `.squad/decisions.md`
- `.squad/decisions/inbox/**`
- `.squad/agents/*/history.md`
- `.squad/casting/*.json`
- `.squad/identity/*.md`
- `.squad/memory/**`
- `.squad/orchestration-log/**`
- `.squad/log/**`
- `.squad/rai/audit-trail.md`
- `.squad/fact-checker/audit-trail.md`

Clarified the local-backend carve-out so `STATE_BACKEND ∈ {"local", "worktree"}` users still freely use `create`/`edit`/`write_file` on `.squad/`.

Also clarified that `memory.*` and `squad_state_*` share the same MCP server (they're aliases in the same registry — see `packages/squad-cli/src/cli/commands/state-mcp.ts`) so models stop treating them as separate availability checks.

## Tests

New `test/state-backend-handshake.test.ts` runs against all 4 template mirror targets and asserts (20 tests = 5 assertions × 4 files):

- Mandatory handshake section exists with "every session" + "before any state mutation" timing
- Probe step exists with `squad_state_health` + tool-discovery mechanism
- HALT step exists with restart-CLI + change-stateBackend remediation
- HARD RULE exists listing the forbidden paths + `create`/`edit`/`write_file` tools by name
- Local/worktree carve-out preserved

20/20 pass; `npm run lint` clean.

## What's NOT in scope (filed/tracked separately)

**Server-side fix (auto-preload MCP tools):** The mechanical root cause is a Copilot CLI feature — preload MCP server tools into the model's function list at session start instead of behind lazy discovery. Out of Squad's control. If maintainers agree it's worth pursuing, file as a feature request against `github/copilot-cli`. With this PR's prompt-level enforcement in place, the server-side fix becomes a nice-to-have rather than load-bearing.

**Skill reinforcement (init-mode, agent-conduct):** Could repeat the rules in `.squad/skills/init-mode/SKILL.md` and `.squad/skills/agent-conduct/SKILL.md` as defense-in-depth for users who skip `squad upgrade`. Worth a small follow-up PR; not blocking.

Closes #1305

### DIFF
--- a/.changeset/fix-1305-state-backend-handshake.md
+++ b/.changeset/fix-1305-state-backend-handshake.md
@@ -1,0 +1,67 @@
+---
+"@bradygaster/squad-cli": minor
+"@bradygaster/squad-sdk": minor
+---
+
+Fix #1305: coordinator must probe for `squad_state` / `memory.*` tools before writing state, and hard-refuse writes when the bridge isn't reachable on non-local backends
+
+## Symptom
+
+A coordinator session against a Squad with `stateBackend: "two-layer"` wrote `.squad/decisions.md` + four `.squad/agents/{name}/history.md` files + a `.squad/decisions/inbox/copilot-directive-…md` via raw `create`/`edit` tools — never calling any `squad_state_*` or `memory.*` tool. The pre-commit hook caught the contract violation; the agent treated it as a "git problem" instead of the symptom it was.
+
+## Root cause (two failure modes stacked)
+
+**Mechanical (Copilot CLI):** Copilot CLI loads MCP server tools lazily — they're registered (via `.mcp.json`) but not always advertised in the model's initial function list. Models have to use `tool_search_tool_regex` to find them or guess they exist. The `squad_state` MCP server is correctly registered with `tools: ["*"]` and responds to `initialize`/`tools/list` over stdio — but the model's tool block at session start may not include them. This is a Copilot CLI architectural choice (lazy discovery vs. preload), not something Squad can fix server-side.
+
+**Behavioral (squad.agent.md):** The pre-1305 prompt said *"When memory tools are available, use them before writing durable memory by hand"* and *"If memory tools are not available, fall back to squad_decide or squad_state_write"*. Models read `"available"` as `"listed in my tool prompt"` instead of `"after probing to find out"`. There was no hard refusal clause for the case where the agent is about to violate the state-backend contract.
+
+## Fix
+
+Two changes to `.squad-templates/squad.agent.md` (synced to all 4 mirror targets):
+
+### 1. New "State-backend handshake" section (MANDATORY, runs once per session)
+
+Inserted right after the `stateBackend` resolution at L131. Steps:
+
+1. If `STATE_BACKEND ∈ {"local", "worktree"}`: file ops on `.squad/` are valid; skip the probe.
+2. Otherwise: probe for `squad_state_health` via `tool_search_tool_regex` (or equivalent tool-discovery mechanism). On success, call `squad_state_health` once to confirm the bridge answers.
+3. **If the probe fails**: HALT before any state write. Output a precise error to the user (verbatim text in the template): *"Squad's runtime state bridge is missing for backend `{STATE_BACKEND}`. The `squad_state` MCP server in `.mcp.json` is not reachable in this Copilot session. Restart Copilot CLI so `.mcp.json` is loaded, or change `stateBackend` to `local` in `.squad/config.json`."*
+
+### 2. Replaced soft "if not available" language with a HARD RULE in `### Memory Governance Tools`
+
+Lists the runtime-owned paths that are FORBIDDEN to write via `create`/`edit`/`write_file` on non-local backends when the bridge isn't reachable:
+
+- `.squad/decisions.md`
+- `.squad/decisions/inbox/**`
+- `.squad/agents/*/history.md`
+- `.squad/casting/*.json`
+- `.squad/identity/*.md`
+- `.squad/memory/**`
+- `.squad/orchestration-log/**`
+- `.squad/log/**`
+- `.squad/rai/audit-trail.md`
+- `.squad/fact-checker/audit-trail.md`
+
+Clarified the local-backend carve-out so `STATE_BACKEND ∈ {"local", "worktree"}` users still freely use `create`/`edit`/`write_file` on `.squad/`.
+
+Also clarified that `memory.*` and `squad_state_*` share the same MCP server (they're aliases in the same registry — see `packages/squad-cli/src/cli/commands/state-mcp.ts`) so models stop treating them as separate availability checks.
+
+## Tests
+
+New `test/state-backend-handshake.test.ts` runs against all 4 template mirror targets and asserts (20 tests = 5 assertions × 4 files):
+
+- Mandatory handshake section exists with "every session" + "before any state mutation" timing
+- Probe step exists with `squad_state_health` + tool-discovery mechanism
+- HALT step exists with restart-CLI + change-stateBackend remediation
+- HARD RULE exists listing the forbidden paths + `create`/`edit`/`write_file` tools by name
+- Local/worktree carve-out preserved
+
+20/20 pass; `npm run lint` clean.
+
+## What's NOT in scope (filed/tracked separately)
+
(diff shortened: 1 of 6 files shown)


tamirdresher:
Reviewer thread addressed in commit 352bb3b5:

- **Regex precedence bug** — replaced the single `/create.*edit.*write_file|create\s*\/\s*edit/i` regex (whose `|` precedence let the shorter `create / edit` branch match without `write_file`) with three separate `\b`-anchored `expect(content).toMatch()` calls, one per tool. Now dropping any one tool name fails the test.

20/20 handshake tests still pass.

bradygaster (approved):
✅ Flight approves. State-backend handshake rule is explicit, the non-local write ban is correctly scoped, template mirrors are synced, and checks are green.

PR #1304, opened 2026-06-13
fix(skills): install bundled skills to .github/skills/ (closes #1126)

tamirdresher:
---
"@bradygaster/squad-sdk": minor
"@bradygaster/squad-cli": minor
---

Move bundled skills from `.copilot/skills/` to `.github/skills/` so they're visible to all Copilot surfaces (closes #1126)

## Symptom (per #1126)

Squad-bundled skills installed at `.copilot/skills/` are **invisible to every Copilot surface except Squad itself**:
- ❌ GitHub Copilot cloud agent
- ❌ Copilot CLI (outside Squad sessions)
- ❌ VS Code Copilot extension (agent mode)
- ❌ `@copilot` coding agent on issues
- ❌ Any future Copilot surface

Per the [official Agent Skills docs](https://docs.github.com/en/copilot/concepts/agents/about-agent-skills), [add-skills docs](https://docs.github.com/en/copilot/how-tos/use-copilot-agents/cloud-agent/add-skills), and [VS Code docs](https://code.visualstudio.com/docs/copilot/customization/agent-skills), the canonical **project-level** custom-skills location is `.github/skills/`. `.copilot/skills/` at repo root is not recognized by any Copilot surface — the home-directory equivalent `~/.copilot/skills/` IS recognized for **personal** skills, which is the source of the original mistake.

## Fix

1. **`squad init`** writes bundled skills to `.github/skills/{name}/SKILL.md` (was `.copilot/skills/`).
2. **`squad upgrade`** does the same AND auto-migrates legacy `.copilot/skills/{manifest-skill}/` into `.github/skills/{manifest-skill}/` (best-effort, preserves user-added non-manifest skills at `.copilot/skills/`, tombstones the legacy dir when empty).
3. **`TEMPLATE_MANIFEST`** destinations rewritten: all 10 skill entries now target `../.github/skills/` instead of `../.copilot/skills/`.
4. **`ENSURE_DIRECTORIES`** (upgrade.ts) updated so existing squads get `.github/skills/` created on upgrade.
5. **squad.agent.md** narrative updated: 5-path scan order now lists `.github/skills/` as path #2 (Copilot CLI's canonical custom-skills location) and `.copilot/skills/` as path #3 (Legacy install path; `squad upgrade` migrates). Personal scope (`~/.copilot/skills/`) preserved as-is.
[...]

PR #1303, opened 2026-06-13
feat(skills): make /squad a real slash command (user-invocable: true + rename squad-commands → squad)

tamirdresher:
---
"@bradygaster/squad-sdk": minor
"@bradygaster/squad-cli": minor
---

Make `/squad` a real slash command — rename `squad-commands` skill to `squad` with `user-invocable: true`

## What this enables

Users can type `/squad` in any Copilot CLI session in a squad-initialized project and get the categorized command catalog (Install & Upgrade, Team Management, Issues & PRs, Plugins & Skills, Model & Cost, Sessions & State). This is what users naturally expect — Squad ships an in-chat command surface, and the slash key is the muscle memory.

## Verified mechanism (Copilot CLI 1.0.62-2)

Decompiling `~/.copilot/pkg/win32-x64/1.0.62-2/sdk/index.js` line 2618:

```js
getLoadedSkills().filter(e => e.userInvocable)
  .map(e => ({name: `/${eF(e)}`, isSkill: true, skill: e}))
```

Any skill with frontmatter `user-invocable: true` is auto-registered by Copilot CLI as a slash command at `/<skill-name>`. The built-in `customize-cloud-agent` skill shipped with Copilot CLI uses this exact pattern with `user-invocable: false`; setting it to `true` makes the skill appear in `/skills` AND as a slash keystroke.

The previous skill `squad-commands` had no `user-invocable` field (Copilot CLI defaults to false), so `/squad-commands` did not exist. Users either typed natural language ("squad commands", "what can squad do") or never discovered the menu.

## Changes

1. **Renamed** the `squad-commands` skill to `squad` (canonical source in `.squad/skills/squad/`).
2. **Frontmatter rewritten** to match Copilot CLI's actual schema (verified against decompiled source):
   - `name: squad` (was `squad-commands` — `/squad` is shorter, intuitive, and matches the agent name without colliding because slash commands and skill lookups are separate namespaces)
   - `user-invocable: true` ← **the load-bearing change**
   - `description:` rewritten to be self-explanatory so natural-language match also still works
   - `allowedTools: []` (matches Copilot CLI's schema)
   - Removed unused fields: `domain:`, `confidence:`, `source:`, `triggers:` (all silently ignored by Copilot CLI)
3. **Body text** updated to reference `/squad` as the primary invocation path.
4. **`MANIFEST_SKILL_NAMES`** in `packages/squad-sdk/src/config/init.ts`: `'squad-commands'` → `'squad'`.
[...]

PR #1301, opened 2026-06-13
fix(sdk): plumb Fact Checker like Rai — rich charter at init + .squad/fact-checker/ state dir (#1299 deep)

tamirdresher:
---
"@bradygaster/squad-cli": minor
"@bradygaster/squad-sdk": minor
---

Fix #1299 (deep): Fact Checker gets the same plumbing as Rai — rich charter at init + state directory + policy template

PR #1300 (also #1299) fixed the documentation gap so the coordinator knows to roster Fact Checker. This PR fixes the **structural** gap behind it. Per user testing on 2026-06-13, even after #1300 the actual agent on disk is still a "name on disk with a 21-line placeholder" — three concrete problems:

| Piece | Rai before this PR | Fact Checker before this PR |
|-------|--------------------|------------------------------|
| `charter.md` at init | Generic 478-byte stub from `generateCharter()` | Generic 523-byte stub from `generateCharter()` |
| Rich charter template usage | Only used by `squad upgrade` (never by `squad init`) | Same — never used by `squad init` |
| `.squad/{name}/policy.md` | ✅ Seeded from `rai-policy.md` (4160 bytes) | ❌ Directory does not exist |
| `.squad/{name}/audit-trail.md` | ✅ Seeded with append-only header | ❌ Directory does not exist |
| `merge=union` in `.gitattributes` | ✅ `.squad/rai/audit-trail.md` | ❌ No entry for fact-checker |
| `fact-checker-charter.md` distribution | n/a | Only in `packages/squad-cli/templates/` — missing from `.squad-templates/` (canonical source) AND `packages/squad-sdk/templates/`, so the SDK init path could never find it |

### Part 1 — Rich charter at init (benefits BOTH Rai and Fact Checker)

`packages/squad-sdk/src/config/init.ts` agent loop now looks up `{templatesDir}/{role}-charter.md` for each agent and uses that as `charter.md` content if it exists. Falls back to `generateCharter()` for user-defined agents that have no rich template (everyone except the built-ins).

Result: a fresh `squad init` produces `.squad/agents/Rai/charter.md` with the full Rai charter (4525 bytes) and `.squad/agents/fact-checker/charter.md` with the full Fact Checker charter (3024 bytes). Previously both were 478-byte stubs.

### Part 2 — `.squad/fact-checker/` state dir, mirroring `.squad/rai/`

Added a new block in `init.ts` (right after the Rai seeding at lines 879–941) that creates:

[...]

PR #1300, opened 2026-06-13
fix(docs): tell coordinator to roster Fact Checker on first-time cast (#1299)

tamirdresher:
Fixes #1299.

## Symptom

When a user runs `squad init` followed by `copilot --agent squad`, the coordinator's first-time casting flow correctly creates `.squad/agents/fact-checker/` on disk (per merged PR #1223) but **omits Fact Checker from the `## Members` table** in `.squad/team.md`. Rai is included, Ralph is included, Scribe is included — Fact Checker is silently dropped.

Verified live on 2026-06-13 in `C:\Temp\test-weekend\` — fresh init produced 8 agent dirs (including fact-checker) but the cast wrote a 7-member team.md (no fact-checker).

## Root cause

`.squad-templates/squad.agent.md` (the canonical coordinator instructions, mirrored to all package templates) had two gaps:

1. **Line 56** said *"Determine team size (typically 4–5 + Scribe)"* — naming only Scribe among the always-on built-ins.
2. The Rai section (line 884) explicitly told the coordinator *"Rai always appears in team.md: `| Rai | RAI Reviewer | ... | 🛡️ RAI |`"* — but **no equivalent section existed for Fact Checker**.

The model added Rai correctly (saw the roster-entry instruction) but had no instruction to add Fact Checker, even though the directory was already scaffolded.

## Fix

1. Updated the team-size line to name all 4 always-on built-ins (Scribe + Ralph + Rai + Fact Checker).
2. Added a full `## Fact Checker — Verification & Devil's Advocate` section that mirrors the Rai pattern:
   - Declares the single-agent dual-mode design (per #789 + #1254) — *"single agent, two modes"*
   - Explicit **Roster Entry** line: `Fact Checker always appears in team.md: \| Fact Checker | Fact Checker | .squad/agents/fact-checker/charter.md | 🔍 Verifier |\`
   - Trigger phrase tables for both Verification mode and Devil's Advocate mode
   - Confidence rating taxonomy (✅/⚠️/❌/🔍)
   - DA brief structure (steelman → assumptions → pre-mortem → alternatives → risk acceptance)
   - Boundaries, background-mode default, state location

## Sync targets

`sync-templates.mjs --sync` propagates `squad.agent.md` to all 4 mirror targets:
- `templates/squad.agent.md.template`
- `packages/squad-cli/templates/squad.agent.md.template`
- `packages/squad-sdk/templates/squad.agent.md.template`
- `.github/agents/squad.agent.md` (this repo's own coordinator file)

[...]

PR #1307, opened 2026-06-13
fix(coordinator): route "spawn a squad" / "another squad" prompts through the cross-squad skill

tamirdresher:
## Symptom

A coordinator initialised by `squad init` saw prompts like *""spawn two squads of designers and devs""* and fanned out raw `task` agents inside its own context, treating ""squad"" as generic English for ""team / group"". It never invoked the bundled `cross-squad` or `cross-squad-communication` skills, so the peer-squad delegation protocol (registry / manifest / sync CLI / git-async / GitHub-issue patterns) was bypassed entirely.

Two structural holes in `squad.agent.md` allowed this:

1. The Routing table had no row that mapped ""spawn a squad"" phrasing to the Squad-PRODUCT concept (only ""upgrade squad"" / ""squad commands"" rows covered Squad-as-a-product vocabulary).
2. The Skill-aware-routing block was process discipline (""check skill directories by domain relevance"") with no hard ""if the user's word matches a skill name, MUST load the skill"" trigger.

## Fix

Two surgical edits to the canonical `.squad-templates/squad.agent.md` (synced to all 4 mirror copies via `scripts/sync-templates.mjs --sync`):

1. **New routing-table row** — matches ""spawn a squad"", ""another squad"", ""two squads"", ""second squad"", ""fan out to squads"", ""delegate to a squad"", or any phrasing that treats ""squad"" as a unit to spawn or address. Action: invoke the `skill` tool on `cross-squad` AND `cross-squad-communication` BEFORE any `task` spawn, then delegate via Pattern 0/1/2/3 — never fan out raw `task` agents in the coordinator's own context.

2. **New ""Hard trigger — keyword-to-skill match"" paragraph** at the top of the Skill-aware-routing block — if any word in the request matches an installed skill name (e.g., ""squad"" → `cross-squad`, ""reflect"" → `reflect`, ""ceremony"" → matching ceremony skill, ""fact-check"" → `fact-checking`, ""release"" → `release-process`), the coordinator MUST invoke the `skill` tool to fully load that skill BEFORE designing its approach. Includes a one-line ""failure mode this rule closes"" pointer so the guard survives future paraphrasing.

3. **Companion `cross-squad/SKILL.md` opener strengthened** — added a one-line ""Read this FIRST any time the user says 'squad' as a thing to spawn / delegate to / address"" callout above the existing `## Context` paragraph, so even a coordinator that skips the routing-table row still hits the trigger when it does eventually load the skill.

## Regression test

`test/template-sync.test.ts` now asserts, for every mirrored copy of `squad.agent.md` (5 locations — canonical `.squad-templates/` + 3 template mirrors + `.github/agents/`), that:

* The routing-table row exists and contains all three trigger phrases (""spawn a squad"", ""another squad"", ""two squads"").
* The row's action cell references the `cross-squad` skill AND the `skill` tool.
* The ""Hard trigger"" + ""keyword-to-skill match"" markers are both present in the Skill-aware-routing block.
* The hard-trigger paragraph names the worked example `""squad"" → cross-squad` so future edits cannot drop the concrete mapping.

**Verified: 223/223 template-sync tests pass.**

## Composability

Disjoint from all other open PRs in the v0.10 stabilisation set (#1292, #1293, #1295, #1298, #1300, #1301, #1302, #1303, #1304, #1306). Only modifies `.squad-templates/squad.agent.md` + its 4 mirrors + `cross-squad/SKILL.md` + the template-sync test.

## Rollout for existing installs

Either:
- `squad upgrade` once this lands (pulls the new template into `.squad/squad.agent.md` automatically), **or**
- Hand-patch `.squad/squad.agent.md` against the diff and restart the Copilot CLI session.

In both cases the user must **restart the Copilot CLI session** for the new routing to take effect — `squad.agent.md` is read once at session start.


bradygaster (approved):
✅ Flight approves. Cross-squad routing now hard-triggers the correct skills, ambiguity is surfaced instead of silently downgraded, mirrors are synced, and regression coverage is in place.

PR #1276, opened 2026-06-11
docs: add feature pages for remaining 8 v0.10 features (tiered-memory, reflect, error-recovery, teams-comms, fleet-dispatch, mcp-frontmatter, dual-mode, skill-security-scanner) + export-import --repo/--branch

tamirdresher:
## What

Closes the remaining 8 (of 11) user-facing v0.10 features from [#1272](https://github.com/bradygaster/squad/issues/1272). Continuation of [#1274](https://github.com/bradygaster/squad/pull/1274) (preset, cross-squad, coordinator-as-agent).

### New pages (8)

| File | Feature | Source verified against |
|------|---------|------------------------|
| `features/tiered-memory.md` | Hot/cold/wiki agent memory model (20-55% context reduction) | `packages/squad-cli/templates/skills/tiered-memory/SKILL.md` |
| `features/reflect.md` | In-session learning skill (HIGH/MED/LOW patterns) | `packages/squad-cli/templates/skills/reflect/SKILL.md` |
| `features/error-recovery.md` | 5 standard failure recovery patterns | `packages/squad-cli/templates/skills/error-recovery/SKILL.md` |
| `features/teams-comms.md` | Microsoft Teams comm adapter (PKCE / device code) | `packages/squad-sdk/src/platform/comms-teams.ts` + changeset |
| `features/fleet-dispatch.md` | `/fleet` hybrid dispatch mode for `squad watch` | `packages/squad-cli/src/cli/commands/watch/capabilities/fleet-dispatch.ts` |
| `features/mcp-frontmatter.md` | `squad init --mcp-frontmatter` flag | `packages/squad-cli/src/cli-entry.ts:346` |
| `features/dual-mode-deployment.md` | `SQUAD_POD_ID` + pod-specific capabilities | `packages/squad-sdk/src/ralph/capabilities.ts` |
| `features/skill-security-scanner.md` | Markdown-aware skill security scanner | `scripts/security-review.mjs` + changeset |

### Edited (1)

| File | What changed |
|------|--------------|
| `features/export-import.md` | Added `--repo` / `--branch` sections for both `squad export` and `squad import` |

## Why

[#1272](https://github.com/bradygaster/squad/issues/1272) identified that ~37% of v0.10 user-facing features had no dedicated docs. [PR #1274](https://github.com/bradygaster/squad/pull/1274) covered the top 3 priorities; this PR closes the remaining 8.

## Source verification

Each page was written against the actual implementation:
- Read each feature's source file before drafting
[...]

PR #1312, opened 2026-06-14
fix(docs): rephrase backtick-in-table cell that breaks docs-build CI on dev

tamirdresher:
Fixes the `Squad CI / test` job that's red on `dev` right now (run [27488323855](https://github.com/bradygaster/squad/actions/runs/27488323855), commit `567f4475`).

## Root cause

The `skill-security-scanner.md` doc landed via #1276 with a 4-backtick table cell intended to display a literal triple-backtick:

```markdown
| Inside a fenced code block (```` ``` ````) | Suppressed |
```

That made `test/docs-build.test.ts`'s regex `/```/g` see 5 occurrences across the file (the real `bash` example contributes 2; the 4-backtick cell's inner `` ``` `` contributes 1; the surrounding `` ```` `` contributes 2 split across opening + closing) — odd total, even-fence check fails. The same construct also tripped the "code blocks contain language specification or valid content" check.

## Fix

Two-line rephrase in `skill-security-scanner.md`: say *"three backticks"* / *"single backtick"* in prose instead of trying to render literal delimiters in a markdown table cell. Reads cleaner anyway.

## Verified locally

- `npx vitest run test/docs-build.test.ts` → 22/22 pass
- `npm run build` in `docs/` → 171 files emitted; pagefind indexes 168 pages / 6911 words

### DIFF
--- a/docs/src/content/docs/features/skill-security-scanner.md
+++ b/docs/src/content/docs/features/skill-security-scanner.md
@@ -66,8 +66,8 @@ The scanner is markdown-aware, which means it understands when a "dangerous" pat
 
 | Where pattern appears | Action |
 |----------------------|--------|
-| Inside a fenced code block (```` ``` ````) | **Suppressed** — treated as documentation, not advice |
-| Inside an inline code span (`` ` ``) | **Suppressed** — treated as a reference |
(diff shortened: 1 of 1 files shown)


bradygaster (approved):
Reviewed: safe quick-win. CI green. LGTM.

PR #1378, opened 2026-06-24
fix(coordinator): restore always-on inline-dispatch gate

tamirdresher:
## Summary

Restores the **always-on inline-dispatch gate** in the Squad coordinator prompt. Since v0.10.0, the coordinator has been **self-executing domain work** (writing code, generating artifacts) instead of dispatching to roster agents — the exact behavior the framework is built to prevent. This was reported by multiple users (coordinator "doing the work itself"), with a community workaround of hand-editing `.github/agents/squad.agent.md`.

Worked in **v0.9.4**, regressed in **v0.10.0**.

## Root cause

Commit `afe78188` — *"fix: context overflow sentinel and coordinator size reduction (#1035)"* — shrank the coordinator template by relocating dispatch **mechanics** into lazy-loaded reference files. That part was correct. But it also **softened the inline-dispatch rule** itself down to a single soft line ("inline work is last-resort fallback only"), which the model reads as a sanctioned *peer escape hatch* rather than a hard constraint.

- This is a **prompt-prose regression, not a code regression** — the SDK coordinator code is byte-identical between the two tags.
- Most pronounced on **VS Code**, where the `runSubagent` mechanics were moved out of the always-on prompt entirely, leaving the model with no in-view dispatch path.

## The fix

Three edits to the canonical source (`.squad-templates/squad.agent.md`), then `npm run sync-templates` to propagate to all 5 copies:

1. **Client Compatibility** — replaced the soft "last-resort fallback" line with an explicit **inline-dispatch gate**: inline is permitted ONLY in Direct Mode, or when *neither* `task` nor `runSubagent` is available. "It's a small task" is explicitly **not** an exemption.
2. **How to Spawn an Agent** — added a one-line **STOP gate**: about to emit a domain artifact with no `task`/`runSubagent` call this turn → STOP and dispatch.
3. Re-inlined the ~6-line **VS Code `runSubagent` micro-playbook** so the dispatch path is always in view on that surface.

Net **+3 lines** to the canonical template (814 → 817).

## Validation

- ✅ New regression test `test/coordinator-inline-dispatch-gate.test.ts` — **red 8/8 → green 8/8**. Asserts the gate anchor, the `STOP gate:` line, and the `runSubagent` micro-playbook across the canonical source **and all 5 synced copies** (parity).
- ✅ `npm run lint` clean (tsc `--noEmit`, both packages).
- ✅ `changelog-gate` green — changeset included (`.changeset/fix-coordinator-inline-dispatch-gate.md`, `patch` for `@bradygaster/squad-cli` + `@bradygaster/squad-sdk`).
- ✅ Independent Lead review: **APPROVE**.

## Shrink goal still holds (#1035)

The #1035 size-reduction goal is preserved — the fix adds only 3 lines:

| Point | Canonical lines |
|---|---|
| v0.9.4 (pre-shrink) | 970 |
| v0.10.0 / #1035 | 723 |
| `dev` today | 814 |
| **this PR (fix)** | **817** |

Final is ~153 lines under pre-shrink — preserves ~94% of #1035's reduction.

## Note on the `.github/agents/squad.agent.md` diff size

The diff for `.github/agents/squad.agent.md` is **larger than the other four mirrors** (~+24 net / 47-line churn vs ~8 lines each). This is expected: the installed copy on `dev` had **pre-existing canonical↔installed drift** (~21 lines behind the canonical source). `sync-templates` reconciled that drift **on top of** the 3-line fix. So that file's churn = 3 (fix) + 21 (drift repair). The other four mirrors were already in sync, so they show only the clean 3-line fix.

## ⚠️ After merge

- This changes the coordinator template, so **existing installs need `squad upgrade`** to pick up the gate.
- Any session currently running on the old coordinator prompt should **restart** to load the corrected behavior.

References the inline-dispatch regression introduced in `afe78188` (#1035).

### DIFF
--- a/.changeset/fix-coordinator-inline-dispatch-gate.md
+++ b/.changeset/fix-coordinator-inline-dispatch-gate.md
@@ -1,0 +1,24 @@
+---
+"@bradygaster/squad-cli": patch
+"@bradygaster/squad-sdk": patch
+---
+
+Fix coordinator inline-dispatch regression: restore always-on dispatch gate in the Squad coordinator template
+
+**Problem**
+
+In v0.10.0 the Squad coordinator started doing domain work itself instead of dispatching to roster agents (it worked in v0.9.4). Root cause: commit `afe78188` (#1035, "context overflow sentinel + coordinator size reduction") relocated the concrete inline-dispatch gate and the dispatch mechanics out of the always-on coordinator prompt (`.squad-templates/squad.agent.md`) into lazy-loaded reference files (`client-compatibility-reference.md`, `spawn-reference.md`). The remaining inline one-liner ("inline work is last-resort fallback only") was too soft, and the concrete "when may I work inline?" rule + the VS Code `runSubagent` how-to were no longer in-context, so the coordinator defaulted to executing work itself.
+
+**Fix**
+
+Three minimal edits to the canonical coordinator template (`.squad-templates/squad.agent.md`), synced to all mirrors via `npm run sync-templates`:
+
+1. **Client Compatibility** — replaced the soft one-liner with an explicit **Inline-dispatch gate**: doing domain work inline is permitted ONLY in Direct Mode, or when NEITHER `task` NOR `runSubagent` is available; otherwise the coordinator MUST dispatch.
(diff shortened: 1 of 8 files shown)


bradygaster (approved):

bradygaster (approved):
Reviewed the dispatch gate fix — surgical, well-tested, addresses a real user-reported regression. The STOP gate and VS Code micro-playbook re-inline are the right call. LGTM.

PR #1391, opened 2026-06-25
feat: log each project at squad init and add `squad projects` command

barichter:
### What

Adds a global, machine-wide registry of every repository where `squad init`
runs, plus a read-only `squad projects` command to list them.

- `squad init` (non-global) now records the project in a global
  `projects.json` so you can later see all of your Squad projects and where
  each one lives on disk.
- `squad projects` lists registered projects, newest first, showing name,
  path, and when each was first registered.

### Why

Once you run Squad across several repos there is no way to recall what you
have initialized or where those projects live. This adds a lightweight,
purely additive memory of your projects. It never changes how a project is
initialized or how its own `.squad/` state behaves.

### How

- New `packages/squad-sdk/src/projects-registry.ts`:
  - `ProjectRegistryEntry { name, path, created_at }`
  - `readProjectsRegistry()`: safe read, returns `[]` on a missing or
    corrupt file so callers never have to guard.
  - `registerProject(name, path)`: idempotent. Re-running for an
    already-registered path updates that entry in place (case-insensitive on
    win32 and darwin, mirroring `FSStorageProvider`) and preserves the
    original `created_at`.
- `squad init` writes to the registry in a best-effort try/catch inside the
  non-global branch, so a registry write failure can never block init.
- `--global` init does NOT register (matches existing global semantics).
- New `squad projects` reader command (table + empty state).

### Scope and safety

- Additive only. No change to existing init behavior or `.squad/` state.
- Registry file lives alongside the existing global Squad state.
- `readProjectsRegistry()` degrades gracefully on a missing or malformed
  file.

### Tests / build

- `npm run build` clean (tsc).
- Lint clean.
- `command-help` test updated; suite green.
- Verified end-to-end against the built CLI in an isolated `APPDATA`
  sandbox: logs at init, `--global` does not log, idempotent re-init, table
  output, and empty-state message all confirmed.

### Files (8 changed, +186/-1)

```
.changeset/log-projects-at-init.md
packages/squad-cli/src/cli-entry.ts
packages/squad-cli/src/cli/commands/projects.ts
packages/squad-cli/src/cli/core/command-help.ts
packages/squad-cli/src/cli/core/init.ts
packages/squad-sdk/src/index.ts
packages/squad-sdk/src/projects-registry.ts
test/cli/command-help.test.ts
```

### Changeset

`@bradygaster/squad-cli` minor, `@bradygaster/squad-sdk` minor.

PR #1392, opened 2026-06-25
feat: add project-navigation commands (open, pick, list, add)

barichter:
> Depends on #1391. Until that PR merges, this PR's diff also includes its single commit (b186993). Both target `dev`; no retarget needed after PR1 merges.

### What

PR1 added a global registry of every project where `squad init` runs, plus a
read-only `squad projects` list. This PR makes that registry actionable. It adds
four navigation commands so you can jump to, choose, alias-list, and register
your Squad projects without leaving the terminal.

**`squad open [name]`**

Resolves a registered project and launches Copilot in that project's directory.

- `squad open <name>`: looks up the project by name, partial name, or path, then
  starts `copilot` in that directory.
- `squad open` with no name, on a TTY, shows a numbered picker.
- `squad open` with no name, when stdout is not a TTY, prints the project list
  and a hint to pass a name, rather than blocking on a prompt.
- `squad open <name> --print-path` (`-p`) prints the resolved path instead of
  launching, for shell integrations like `cd (squad open foo -p)`.

**`squad pick`**

An always-interactive opener. It presents the numbered picker of all registered
projects, prompts for a choice, then opens the chosen project in Copilot. It is
equivalent to running `squad open` after manually finding a name in the list.
Positional arguments are stripped so it always enters the picker; flags such as
`--print-path` are forwarded.

**`squad list`**

A convenience alias for `squad projects`. It accepts the same flags, including
`--open`, and behaves identically. Existing `squad projects` usage is unchanged.

**`squad add [path]`**

Registers a pre-existing directory in the global registry so it shows up in
`squad projects`, `squad list`, and `squad pick`, without re-running
`squad init` and without scaffolding a `.squad/` directory.

- `squad add <path>`: registers the directory at that path. With no path, it
  registers the current directory.
- `squad add <path> --name <name>`: sets a custom display name. Without it, the
  directory basename is used.
- Paths with spaces are handled even when passed unquoted: the remaining
  argument tokens are rejoined into the full path. If the reconstructed path does
  not exist and there were multiple tokens, it prints a hint to quote the path.

### Why

A list of projects is only useful if you can act on it. These commands close the
loop: from "what have I initialized?" to "take me there," to "let me choose from
a menu," to "remember this directory too." `squad open` is the core verb,
`squad pick` is the discoverable interactive form, `squad list` is muscle-memory
ergonomics, and `squad add` lets you backfill projects that predate the registry
or that you set up by hand.

### How

**SDK resolver: `resolveProject(query)`** in `projects-registry.ts`. Resolution
order, first match wins:

1. Exact name match (case-insensitive on win32 and darwin). One hit returns
   `{ match }`; several with the same name return `{ ambiguous }`.
2. Exact path match via `samePath` on the resolved absolute paths.
3. Unique case-insensitive substring match on name. One hit returns `{ match }`,
   several return `{ ambiguous }`.

It returns a discriminated union: `{ match }`, `{ ambiguous: [...] }`, or
`{ notFound: true }`. An empty query (after trimming) always returns
`{ notFound: true }`.

**SDK registration: `registerProject(name, path)`** (introduced in PR1, reused
here by `squad add`). It is idempotent: re-registering a path that already exists
updates that entry in place and preserves the original `created_at`, rather than
adding a duplicate. `squad add` first checks whether the path is already present
so it can report "Added" versus "Updated" accurately.

**Command implementations:**

- `open.ts` owns all shared opener logic: the resolver call, the empty-registry
  and not-found and ambiguous messages, the numbered picker, the non-TTY list
  fallback, the missing-on-disk guard, `--print-path`, and the Copilot launch.
  Launch is platform-aware (`shell: true` on win32) and inherits stdio. If
  `copilot` is not on PATH (ENOENT), it prints the path with a hint instead of
  crashing.
- `pick.ts` delegates to `runOpen` with positional args removed, so all picker
  and launch behavior is reused with no duplication.
- `projects.ts` gains an `--open`/`-o` branch that delegates to `runOpen`;
  listing remains the default.
- `add.ts` parses `--name`, rejoins remaining tokens into a path, resolves it to
  absolute, validates it exists and is a directory, then calls `registerProject`.

`cli-entry.ts` wires `open`, `pick`, and `add` to their handlers and routes both
`projects` and `list` to `runProjects`. `command-help.ts` adds dedicated help
blocks for `open`, `pick`, and `add`, updates the `projects` help to document
`--open`, and registers `list` as an alias of `projects`.

### Scope and safety

- `squad open`, `squad pick`, and `squad list` are read-only against the
  registry. They never write.
- `squad add` DOES write to the registry: it calls `registerProject`, which adds
  or updates a single entry in `projects.json`. It does not scaffold `.squad/`,
  does not touch project contents, and does not change how `squad init` behaves.
  Re-adding an existing path updates in place and preserves `created_at`.
- No change to PR1's behavior or to any project's `.squad/` state.
- Every command degrades gracefully on edge cases: empty registry, no match,
  ambiguous match, non-TTY invocation, missing-on-disk path, `copilot` not
  installed, nonexistent or non-directory `add` target, and spaced unquoted
  paths.

### Tests / build

- `npm run build` clean (tsc, both packages).
- Lint clean (`tsc --noEmit` on both package tsconfigs).
- Targeted suites green, 21 tests passing:
  - `test/cli/add.test.ts` (4): registers and shows up in `squad list`, clear
    error for a nonexistent path, spaced-path reconstruction from split argv
    tokens, and the quoting hint for a nonexistent multi-token path. All assert
    no stack trace leaks.
  - `test/cli/command-help.test.ts` (14): the command-help end-to-end suite,
    covering the help blocks and the command list now including the new verbs.
  - `test/global-squad-isolation.test.ts` (3): proves the new test harness
    redirects the global Squad path to an OS temp dir.
- New test-isolation infrastructure: `test/setup/isolate-global-squad.ts`
  redirects `resolveGlobalSquadPath()` to a per-worker temp dir (via `APPDATA`,
  `LOCALAPPDATA`, `XDG_CONFIG_HOME`, and `HOME`), wired through `vitest.config.ts`
  as a setup file. Because PR1 made `squad init` write to the global
  `projects.json`, this prevents the test suite from polluting a developer's real
  registry.

### Files (17 changed, +590/-4)

```
.changeset/squad-add-command.md
.changeset/squad-list-alias.md
.changeset/squad-open-command.md
.changeset/squad-pick-command.md
packages/squad-cli/src/cli-entry.ts
packages/squad-cli/src/cli/commands/add.ts
packages/squad-cli/src/cli/commands/open.ts
packages/squad-cli/src/cli/commands/pick.ts
packages/squad-cli/src/cli/commands/projects.ts
packages/squad-cli/src/cli/core/command-help.ts
packages/squad-sdk/src/index.ts
packages/squad-sdk/src/projects-registry.ts
test/cli/add.test.ts
test/cli/command-help.test.ts
test/global-squad-isolation.test.ts
test/setup/isolate-global-squad.ts
vitest.config.ts
```

### Changeset

Four changeset files, all minor:

- `squad-open-command.md`: `@bradygaster/squad-cli` minor, `@bradygaster/squad-sdk` minor.
- `squad-pick-command.md`: `@bradygaster/squad-cli` minor.
- `squad-list-alias.md`: `@bradygaster/squad-cli` minor.
- `squad-add-command.md`: `@bradygaster/squad-cli` minor.

The SDK bump rides on the open/resolver changeset since `resolveProject` is a new
public SDK export; the other three are CLI-only.

### DIFF
--- a/.changeset/log-projects-at-init.md
+++ b/.changeset/log-projects-at-init.md
@@ -1,0 +1,15 @@
+---
+"@bradygaster/squad-cli": minor
+"@bradygaster/squad-sdk": minor
+---
+
+Log each project at `squad init` and add `squad projects` to list them
+
+`squad init` now records the project (name, absolute path, and creation timestamp) in a global registry, `projects.json`, stored under the existing global Squad home (`resolveGlobalSquadPath()`). A new `squad projects` command reads that registry and prints every Squad project on the machine, newest first.
+
+This closes a small but common gap: after working across several repositories, a user had no built-in way to remember where all of their Squad projects live. The feature is purely additive:
+
+- **`squad-sdk`** gains `registerProject()` and `readProjectsRegistry()` (plus the `ProjectRegistryEntry` type). The registry write is idempotent on the project path (re-running `squad init` updates the entry in place, never duplicating) and fail-safe (a registry error never blocks init).
+- **`squad-cli`** calls `registerProject()` during repo `squad init` (global `--global` init is intentionally not registered, since it is the personal home, not a project) and adds the `squad projects` reader command.
+
+No existing init output, project `.squad/` state, or other command behavior changes.
--- a/.changeset/squad-add-command.md
+++ b/.changeset/squad-add-command.md
@@ -1,0 +1,7 @@
+---
+"@bradygaster/squad-cli": minor
+---
+
+Add `squad add` to register an existing directory as a Squad project.
+
+`squad add <path>` registers a pre-existing directory in the global project registry so it appears in `squad projects`, `squad list`, and `squad pick`, without re-running `squad init`. The command handles paths containing spaces even when passed without quotes, reconstructing the full path from remaining CLI arguments.
--- a/.changeset/squad-list-alias.md
+++ b/.changeset/squad-list-alias.md
@@ -1,0 +1,7 @@
+---
+"@bradygaster/squad-cli": minor
+---
+
+Add `squad list` as an alias for `squad projects`.
+
+`squad list` is a convenience alias: it accepts the same flags as `squad projects` (including `--open`) and behaves identically. Existing `squad projects` usage is unchanged.
--- a/.changeset/squad-open-command.md
+++ b/.changeset/squad-open-command.md
@@ -4,3 +4,5 @@
 ---
 
 Add `squad open` to launch a registered project in Copilot.
+
+- `squad projects` now accepts `--open [name]` (alias `-o`) to launch a project from the list, delegating to `squad open`. Listing remains the default behavior.
--- a/.changeset/squad-open-command.md
+++ b/.changeset/squad-open-command.md
@@ -1,0 +1,6 @@
+---
+"@bradygaster/squad-cli": minor
+"@bradygaster/squad-sdk": minor
+---
+
+Add `squad open` to launch a registered project in Copilot.
--- a/.changeset/squad-pick-command.md
+++ b/.changeset/squad-pick-command.md
@@ -1,0 +1,7 @@
+---
+"@bradygaster/squad-cli": minor
+---
+
+Add `squad pick` as an interactive project picker.
+
+`squad pick` presents a numbered menu of all registered Squad projects, prompts the user to choose one, then opens it in Copilot. It is equivalent to running `squad open` after manually locating a project name in `squad projects`.
--- a/packages/squad-cli/src/cli-entry.ts
+++ b/packages/squad-cli/src/cli-entry.ts
@@ -179,6 +179,7 @@ async function main(): Promise<void> {
     console.log(`             Alias: list`);
     console.log(`  ${BOLD}open [name]${RESET} Open a registered project in Copilot`);
     console.log(`  ${BOLD}pick${RESET}       Show an interactive picker and open the chosen project`);
+    console.log(`  ${BOLD}add [path]${RESET}  Register an existing directory as a Squad project`);
     console.log(`  ${BOLD}roles${RESET}      List built-in Squad roles`);
     console.log(`             Usage: roles [--category <name>] [--search <query>]`);
     console.log(`  ${BOLD}cost${RESET}       Report token usage from orchestration logs`);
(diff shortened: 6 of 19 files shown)


tamirdresher:
✅ Reviewed — clean feature, well-tested. Approving.

🔒 **One security note:** In `open.ts`, `launchCopilot()` uses `spawn('copilot', [], { cwd: entry.path, shell: true })` on Windows. With `shell: true`, `cmd.exe` checks CWD for `.cmd`/`.bat` before PATH — so a malicious `copilot.cmd` planted in a registered project dir would execute instead of the real binary.

**Suggested fix:** Resolve `copilot` to its absolute path before spawning, then use `shell: false`. Or reuse the existing `resolveCopilotCmd()` pattern from `agent-spawn.ts`.

Low urgency (requires attacker to plant file in a cloned repo), but worth a follow-up patch. Ship it as-is if you're ok with it. 🚀

— Tamir (via Squad)

tamirdresher (approved):
Approved. See my comment re: shell:true CWD hijack on Windows — non-blocking, follow-up fix recommended.

tamirdresher:
@bradygaster — checking in on the coupled #1391 / #1392 pair so you can steer next steps.

**Current state**
- Both CONFLICTING against `dev`; #1392 is ~111 commits behind, no activity for 9 days.
- #1392 explicitly depends on #1391 and already contains its commit (`b186993`).
- #1391 has 1 open Copilot review thread (atomic registry write); #1392 has 9 open Copilot threads (arg validation, TTY gating on stdin+stdout, ambiguous-resolution UX including path disambiguator, `fatal()` wrapping in `add.ts`, cross-platform isolation-test assertions, etc.).
- My earlier security note on #1392 — `launchCopilot()` uses `spawn('copilot', [], { shell: true })` and on Windows `cmd.exe` resolves `.cmd`/`.bat` from CWD before PATH, so a planted `copilot.cmd` in a registered project directory would win — is still unaddressed (no fix commit yet).

**Which shape do you want the author to land in?**

1. **Rebase separately** — land #1391 first (registry + `squad projects`), then rebase #1392 on top for the navigation commands (`open`, `pick`, `list`, `add`).
2. **Consolidate into #1392** — close #1391 (its single commit is already inside #1392) and land the full feature as one rebased PR.

Option 2 is fewer round-trips if you're comfortable losing #1391's isolated history; option 1 keeps a cleaner review boundary between the registry and the navigation surface. Your call — leaving one comment here to avoid duplicating on #1391.

bradygaster:
Closing as stale after the approved backlog triage. This PR depends on now-closed #1391, is 145 commits behind dev with merge conflicts, has nine unresolved review threads, and still carries the unaddressed Windows shell/CWD executable-resolution risk. Please resubmit the navigation feature as a fresh PR against current dev, with the registry foundation incorporated or separately reintroduced and the shell:true risk removed.

PR #1406, opened 2026-06-29
fix(cli): pin squad-sdk to workspace:* (fixes prerelease build crash)

### LINKED ISSUE #1405: CLI crashes/build break: squad-cli pins squad-sdk '>=0.9.0' so prerelease builds pull stale published SDK
## Summary
`packages/squad-cli/package.json` depends on `@bradygaster/squad-sdk: ">=0.9.0"`. The workspace is a prerelease (`0.10.0-build.2`), which `>=0.9.0` does NOT satisfy, so `npm install` puts the **published 0.10.0** SDK in `packages/squad-cli/node_modules/@bradygaster/squad-sdk` instead of linking the workspace. That copy lacks new exports → build + runtime crash.

## Crash signature
`SyntaxError: ... does not provide an export named 'addSquadStateGitignoreBlock'` (also `installPresetFromSource`, `readSquadRegistry`, `stateBackend`). `tsc -b` → TS2305 across cross-squad.ts, migrate-backend.ts, preset.ts, init.ts.

## Repro
`npm install && npm run build` → CLI build fails. `rm -rf packages/squad-cli/node_modules/@bradygaster` → builds clean; failures drop 108→13.

## Fix
Pin to workspace: `"@bradygaster/squad-sdk": "workspace:*"`.

## Release impact (0.11.0)
Blocker. Also fails `test/cli-packaging-smoke` (packed CLI grabs published SDK). Pre-existing consult/personal-squad path-drift failures are separate.

bradygaster:
Closes #1405.\n\nWorking as EECOM (Core Dev).\n\nPins the CLI dependency on @bradygaster/squad-sdk to workspace:* so prerelease builds link the local workspace SDK instead of pulling a stale published copy, avoiding the missing-export build/runtime crash.\n\nValidation:\n- SKIP_BUILD_BUMP=1 npm run build

### DIFF
--- a/.changeset/fix-cli-sdk-workspace-pin.md
+++ b/.changeset/fix-cli-sdk-workspace-pin.md
@@ -2,4 +2,4 @@
 "@bradygaster/squad-cli": patch
 ---
 
-Pin CLI dependency on squad-sdk to workspace:* so prerelease builds link the local SDK instead of pulling a stale published copy (fixes build/runtime crash). Closes #1405.
+Force the CLI's squad-sdk dependency through the repo workspace by using a root file dependency plus an npm override for the CLI package. This keeps fresh installs linked to the local SDK prerelease instead of pulling the stale published SDK. Closes #1405.
--- a/.changeset/fix-cli-sdk-workspace-pin.md
+++ b/.changeset/fix-cli-sdk-workspace-pin.md
@@ -1,0 +1,5 @@
+---
+"@bradygaster/squad-cli": patch
+---
+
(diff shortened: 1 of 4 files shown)


tamirdresher (approved):
✅ LGTM — clean fix for the workspace linkage.

**What I verified:**
- Root \ile:packages/squad-sdk\ dep + \overrides\ with \\$\ reference is the documented npm pattern for forcing workspace resolution
- CLI relaxes to \*\ so the override wins without version mismatch
- Lockfile diff is just removal of stale transitive otel deps from the published SDK — expected
- Changeset is correctly scoped to \squad-cli: patch\

The only minor note: root \package.json\ version went from \

Issue #920, opened 2026-04-08
Windows: spawn EINVAL in watch capabilities (execFile without shell:true)


tamirdresher:
### Environment
- OS: Windows 11
- Node: v22+
- squad-cli: `@bradygaster/squad-cli@0.9.4-insider.1`
- agency: 2026.4.8.1

### Problem
Capabilities `monitor-teams`, `monitor-email`, `retro`, and `decision-hygiene` fail with `spawn EINVAL` on Windows when running `watch --execute`.

### Root Cause
These capabilities use `execFile(cmd, args, ...)` without `shell: true`. On Windows, `execFile` cannot:
1. Resolve bare command names (like `gh`) through PATH
2. Execute `.cmd`/`.bat` wrapper scripts

The `execute` capability works fine because it handles this differently.

### Additional Issue
These capabilities hardcode `gh copilot --message` but don't respect the `agentCmd` config. Users with `copilot.exe` (standalone WinGet) or `agency` CLI can't use these capabilities even with a working `agentCmd` configured.

### Suggested Fix
1. Add `shell: true` to the `execFile` options in `spawnWithTimeout()`
2. Have all capabilities respect `agentCmd` from config (not just `execute`)

### Workaround
Disable the broken capabilities in `.squad/config.json`:
```json
{
  "watch": {
    "monitor-teams": false,
    "monitor-email": false,
    "retro": false,
    "decision-hygiene": false
  }
}
```

### Output
```
spawn EINVAL on monitor-teams, monitor-email, retro, decision-hygiene
```

tamirdresher:
**Clarification:** `gh copilot` is the deprecated GH CLI extension. The current tool is standalone `copilot` (installed via WinGet at `copilot.exe`).

The preferred pattern (matching the PS1 ralph-watch) is `agency copilot --yolo --autopilot --mcp mail --agent squad -p <promptfile>`.

So the fix should be:
1. All capabilities should respect `agentCmd` from `.squad/config.json` (not just `execute`)
2. Stop hardcoding `gh copilot` — it's deprecated
3. Add `shell: true` to `execFile()` calls on Windows so `.cmd` wrappers and PATH resolution work

The `agentCmd` config already exists and works for `execute` — just wire it into the other capabilities too.

PR #1437, opened 2026-07-02
refactor(cli): consolidate duplicated agent-spawn logic across watch capabilities

### LINKED ISSUE #994: refactor: consolidate duplicated buildAgentCommand and spawnWithTimeout across watch capabilities
## Context
PR #988 fixed the --message flag across 8 files, but the underlying pattern (same command-building logic copy-pasted everywhere) is still there. Flagged by @diberry in her review.

## What
1. Extract a shared buildAgentCommand() into a single location (e.g., watch/utils.ts)
2. Consolidate spawnWithTimeout() which appears in 4 places with identical implementation
3. Update docs/comments that still reference gh copilot (code now calls copilot directly)

## Why
Prevents this class of bug from recurring. Next time the Copilot CLI changes a flag, it is one file to update instead of 8.

Follow-up from PR #988. Credit: @diberry review.

omercangumus:
## Summary

Closes #994

`execute.ts` and `wave-dispatch.ts` each carried their own copy of `buildAgentCommand()` and an inline spawn-with-timeout implementation. These had drifted from the shared `agent-spawn.ts` module (added for #920/#923) because they needed the `--additional-mcp-config`/`--yolo` workaround from `withAdditionalMcpConfig()` (see `copilot-invocation.ts`) that the shared `buildAgentCommand()` doesn't apply.

## Root cause

- `agent-spawn.ts` exports a canonical `buildAgentCommand()`/`spawnWithTimeout()`/`spawnAgent()` used by `decision-hygiene`, `monitor-email`, `monitor-teams`, and `retro`.
- `execute.ts` and `wave-dispatch.ts` were never migrated to it because they need MCP-config injection that the canonical `buildAgentCommand()` doesn't provide, so each kept its own near-identical copy (and `execute.ts`'s copy also needed to track/untrack the spawned PID via `context.pidTracker`, which the shared `spawnAgent()` didn't support).
- Result: two extra copies of the same command-building/spawn-with-timeout logic, and `execute.ts`'s own inline `execFile` call was missing the `shell: IS_WINDOWS` + argument-escaping fix that was already applied to the shared module in #1388 (Windows DEP0190).

## Change

- Added `buildCopilotCommand()` to `agent-spawn.ts` — the MCP-aware command builder that `execute.ts` and `wave-dispatch.ts` both need.
- Extended `spawnAgent()` with an optional `pidTracking` parameter so `execute.ts` can register/unregister its spawned session with `context.pidTracker` without keeping its own `execFile` call.
- `execute.ts` and `wave-dispatch.ts` now both import from `agent-spawn.ts` instead of defining local copies.
- As a side effect, `execute.ts`'s Copilot session spawn now goes through the same Windows-safe (`shell: true` + escaped args) path as the rest of the watch capabilities.
- Fixed a stale `(default: gh copilot)` diagnostic label in `index.ts`'s verbose startup table — the actual default is the bare `copilot` binary, with `gh copilot` only as a runtime fallback.

## Testing

- `npm run build`, `npm test`, `npm run lint` all pass.
- Added `test/cli/watch-agent-spawn.test.ts` covering `buildCopilotCommand()` (agentCmd override, default flags, MCP-config injection) and `spawnAgent()`'s new pid-tracking behavior.
- Added a regression test in `test/cli/watch-capabilities.test.ts` verifying `ExecuteCapability` still tracks/untracks the spawned PID through `context.pidTracker` after the refactor.
- All existing tests in `test/cli/watch-capabilities.test.ts`, `test/cli/watch-execute.test.ts`, `test/cli/watch-fleet-dispatch.test.ts`, and `test/watch-fleet-dispatch.test.ts` pass unchanged, confirming behavior is preserved.
- Note: a full `npm test` run on this Windows dev box shows some pre-existing, unrelated timeouts/EBUSY failures (e.g. `test/cli/state-mcp.test.ts`, `test/cli/watch.test.ts`, `test/cli/upgrade.test.ts`) under environment load; I confirmed these reproduce identically on a clean `upstream/dev` checkout, so they're unrelated to this change.

### DIFF
--- a/.changeset/consolidate-watch-agent-spawn.md
+++ b/.changeset/consolidate-watch-agent-spawn.md
@@ -1,0 +1,5 @@
+---
+"@bradygaster/squad-cli": patch
+---
+
+Consolidate the duplicated `buildAgentCommand()`/spawn-with-timeout logic in the `execute` and `wave-dispatch` watch capabilities into the shared `agent-spawn.ts` module (#994). As a side effect, `execute`'s Copilot session spawn now goes through the same Windows `cmd.exe` argument-escaping path (DEP0190) as the rest of the watch capabilities.
--- a/packages/squad-cli/src/cli/commands/watch/agent-spawn.ts
+++ b/packages/squad-cli/src/cli/commands/watch/agent-spawn.ts
@@ -11,6 +11,7 @@
 
 import { execFile, execFileSync } from 'node:child_process';
 import type { WatchContext } from './types.js';
+import { withAdditionalMcpConfig } from '../../core/copilot-invocation.js';
 
 /** True when running on Windows — used to gate `shell: true`. */
 export const IS_WINDOWS = process.platform === 'win32';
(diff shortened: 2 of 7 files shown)


tamirdresher (approved):
✅ Approved — security review passed.

**Security checks:**
- No new command injection surface: \gentCmd\ splitting pattern is identical to the original (config-derived, not user-input)
- \prompt\ passed via \-p\ flag — same as before, just centralized
- **Security hardening bonus:** execute.ts now routes through \scapeArgs()\ on Windows (fixes DEP0190 shell injection)
- PID tracking labels derived from issue numbers, not raw user text
- No new env var reads, no new network calls, no new shell: true usage

**Correctness:** Clean deduplication. execute.ts and wave-dispatch.ts now share one code path. Tests verify pid tracking lifecycle (track on spawn, untrack on exit) and the agentCmd/copilotFlags/MCP-injection branches.

Solid refactoring work 🔧

Issue #1190, opened 2026-05-29
bug: upgrade pipeline gaps — postinstall misses repo-local node_modules, two-layer hooks not installed, teamRoot not portable


tamirdresher:
## Summary

After upgrading to `0.9.6-insider.3` with `stateBackend=two-layer`, three related gaps in the upgrade pipeline leave a repo in a subtly broken state that `squad doctor` only partially catches.

## Environment

- squad-cli: `0.9.6-insider.3` (global), `0.9.1` (repo-local)
- Windows 11, PowerShell, Node v23.5.0
- `stateBackend: "two-layer"` in `.squad/config.json`

---

## Finding 1 — `postinstall` ESM patch misses repo-local `node_modules` (references #449)

`squad doctor` reports two failures after a global install of `0.9.6-insider.3`:

```
❌  vscode-jsonrpc exports field — missing exports field — run postinstall or reinstall (see #449)
❌  copilot-sdk session.js ESM patch — session.js has extensionless vscode-jsonrpc/node import
```

Running `npm run postinstall` from the global squad-cli package dir reports **"already patched — skipping"** yet doctor still fails. Root cause: `patch-esm-imports.mjs` uses hardcoded `SEARCH_ROOTS` relative to `__dirname` of the global install:

```js
const SEARCH_ROOTS = [
  join(__dirname, '..', 'node_modules'),              // squad-cli local
  join(__dirname, '..', '..', '..', 'node_modules'),  // workspace root
  join(__dirname, '..', '..'),                         // global install (sibling)
];
```

None of these resolve to `<repo>/node_modules`. The repo-local `node_modules` (where doctor actually checks) is never patched.

**Fix:** Add `join(process.cwd(), 'node_modules')` to `SEARCH_ROOTS`. Additionally, `squad upgrade` should invoke the patch script from the **repo root** (not the global package dir) after install completes.

---

## Finding 2 — `pre-commit` / `post-commit` hooks missing after two-layer upgrade (references #1185)

Because `--state-backend` migration was ignored during upgrade (#1185), the `pre-commit` and `post-commit` hooks required by two-layer were never installed. The `squad-state` orphan branch exists but has only its init commit — nothing has ever written to it.

Installed hooks after upgrade:
| Hook | Status |
|---|---|
| `pre-push` | ✅ |
| `post-checkout` | ✅ |
| `post-merge` | ✅ |
| `post-rewrite` | ✅ |
| `pre-commit` | ❌ Missing |
[...]

Issue #1185, opened 2026-05-28
squad upgrade --self --insider: misplaced templates, Rai not installed, --state-backend ignored


ischrei:
Upgraded from `v0.9.6-insider.2` -> `v0.9.6-insider.3` via `squad upgrade --self --insider`. Three distinct issues:

### 1. Templates dumped at `.squad/` root
~20 template files were copied into `.squad/` root **in addition to** `.squad/templates/` (e.g. `Rai-charter.md`, `rai-policy.md`, `scribe-charter.md`, `charter.md`, `history.md`, `roster.md`, `casting-*.json`, `copilot-instructions.md`, `mcp-config.md`, `multi-agent-format.md`, `orchestration-log.md`, `plugin-marketplace.md`, etc.). All were exact duplicates of `.squad/templates/*` — had to delete manually.

### 2. Rai not actually installed
Despite Rai being a built-in roster member, upgrade did NOT:
- create `.squad/agents/Rai/{charter,history}.md`
- create `.squad/rai/{policy,audit-trail}.md`
- add Rai row to `team.md` or `routing.md`
- add `.squad/rai/audit-trail.md merge=union` to `.gitattributes`

### 3. `--state-backend` flag ignored
From existing `orphan` backend, neither worked:
```
squad upgrade --state-backend two-layer
squad upgrade --state-backend local
```
Got an error that upgrade from "orphan" to "two-layer" isn't supported.

### Environment
- squad-cli: `0.9.6-insider.3` (from `0.9.6-insider.2`), channel: insider
- Windows 11, PowerShell, Node v24.14.1 (nvm)

### Expected
1. Upgrade should not dump templates at `.squad/` root.
2. Built-in roster members (Rai) should auto-install if missing.
3. `--state-backend <value>` should migrate

bradygaster:
Thanks for reporting these — the insider channel has since been eliminated (see #1096) and we're now on v0.10.0 with a simplified branching strategy. These upgrade paths no longer exist. Closing as outdated.

PR #1463, opened 2026-07-10
fix(cli): upgrade patches repo-local node_modules; doctor verifies two-layer commit hooks

omercangumus:
### What
Findings 1 and 2 from #1190: `squad upgrade` now applies the ESM import patch to the repo-local `node_modules`, and `squad doctor` fails when the `pre-commit`/`post-commit` hooks required by the two-layer/orphan backends are missing.

### Why
Part of #1190 — finding 3 (teamRoot portability) lands separately; asked on the issue whether `peers.<machineId>.teamRoot` is the intended convention before building it.

Two gaps left a repo subtly broken after upgrading:

1. npm runs `postinstall` with cwd inside the installed package dir, so a global `npm install -g` only ever patched the global copy of vscode-jsonrpc / copilot-sdk — the consumer repo's `node_modules` stayed unpatched and `squad doctor` kept failing its #449 checks while the patcher reported "already patched — skipping". The `process.cwd()` entry added to `SEARCH_ROOTS` in 2d9f0b4e couldn't help: both patch functions returned at the first root that already had a patched copy, which on a global install is always the global one, so the cwd entry was effectively dead code.
2. `doctor`'s hook check only covered the four sync hooks (`pre-push`/`post-merge`/`post-rewrite`/`post-checkout`). A two-layer repo missing `pre-commit`/`post-commit` — the hooks that actually guard and flush state to the `squad-state` branch — passed the check while the state branch sat dormant, which is the exact post-#1185 condition described in the issue.

### How
- `scripts/patch-esm-imports.mjs`: both patch functions now iterate every search root instead of returning at the first hit, log which path they patched or skipped, and are exported. A direct-invocation guard keeps `node scripts/patch-esm-imports.mjs` (postinstall) behavior unchanged.
- `core/upgrade.ts`: new `ensureEsmImportsPatched(dest)` locates the script by walking up from the compiled file (same approach as `getTemplatesDir()`), imports it, and points it at `<dest>/node_modules`. Called from `runEnsureChecks()`, so it covers both the normal and already-current upgrade paths — rerunning `squad upgrade` on an up-to-date repo still repairs the patch.
- `commands/doctor.ts`: `REQUIRED_SYNC_HOOKS` now includes `pre-commit`/`post-commit`, matching the full set `install-hooks.ts` installs and `ensureHooksForBackend` verifies. The existing fail message already names each missing hook and points at `squad install-hooks`.

### Testing
- New `test/cli/patch-esm-imports.test.ts` (6 tests): the all-roots regression for both layers (an already-patched earlier root no longer masks a later unpatched one), idempotency, and `ensureEsmImportsPatched` wiring against a temp project's node_modules. 
- `test/cli/doctor.test.ts`: new regression test for the exact #1190 state (four sync hooks present, commit hooks missing → fail, message contains both hook names and the remediation); the two existing PASS fixtures updated to install all six hooks.
- `npm run build` — passes. `npm run lint` (tsc) — passes. `npm run lint:eslint` — 0 errors (the new helper carries the same pre-existing n/no-sync warning class the rest of `core/upgrade.ts` already has).
- `npx vitest run test/cli/patch-esm-imports.test.ts test/cli/doctor.test.ts` — 42 passed.
- Full `npm test` on this Windows box: 59 failed / 6962 passed with these changes vs 64 failed / 6950 passed on a clean `upstream/dev` checkout using the same build+run procedure. The failing sets shift run to run and are the known local timing/locale flakes (5s timeouts in journey/e2e/shell tests, a unicode-vs-ascii rendering assertion); every test that failed only in the with-changes run passes standalone against the same build, and 50 of the failures are common to both runs.

---

### ⚠️ Quick Check
- [x] Changeset added: `.changeset/fix-1190-esm-patch-doctor-commit-hooks.md` (patch bump, `@bradygaster/squad-cli` only — no SDK or governed template paths touched)

#### Branch & Commit
- [x] Verified diff contains only intended changes (`git diff --cached --stat`, plus a `git diff -w` pass for whitespace-only hunks — none)
- [x] PR is not in draft mode

#### Build & Test
- [x] `npm run build` passes
- [x] `npm test` passes (see testing notes above for pre-existing unrelated local failures)
- [x] `npm run lint` passes (type check clean)
- [x] `npm run lint:eslint` passes (0 errors)

#### Changeset
- [x] Changeset added (patch, `@bradygaster/squad-cli`)

#### Docs
- [x] N/A — behavior fix inside existing upgrade/doctor flows, no user-facing API change

#### Exports
- [x] N/A — `ensureEsmImportsPatched` is exported from `core/upgrade`, which already has a subpath export

---

### Breaking Changes
None.

### Waivers
None.

### DIFF
--- a/.changeset/fix-1190-esm-patch-doctor-commit-hooks.md
+++ b/.changeset/fix-1190-esm-patch-doctor-commit-hooks.md
@@ -1,0 +1,5 @@
+---
+"@bradygaster/squad-cli": patch
+---
+
+Fix two upgrade-pipeline gaps from #1190: the postinstall ESM patcher now patches every search root instead of stopping at the first already-patched copy, and `squad upgrade` re-runs it against the repo-local node_modules so globally-installed CLIs no longer leave the consumer repo's vscode-jsonrpc/copilot-sdk unpatched. `squad doctor` now also verifies the pre-commit/post-commit hooks required by the two-layer/orphan state backends, instead of only the four sync hooks.
--- a/packages/squad-cli/scripts/patch-esm-imports.mjs
+++ b/packages/squad-cli/scripts/patch-esm-imports.mjs
@@ -18,7 +18,7 @@
 
 import { readFileSync, writeFileSync, existsSync } from 'fs';
 import { join, dirname } from 'path';
-import { fileURLToPath } from 'url';
+import { fileURLToPath, pathToFileURL } from 'url';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 
@@ -36,8 +36,12 @@ const SEARCH_ROOTS = [
  * This is the canonical fix: once the package has proper exports, Node's
  * ESM resolver handles every subpath ('vscode-jsonrpc/node', '/browser', etc.)
  * without needing per-file patches.
+ *
+ * Patches EVERY root that contains a copy, not just the first one found:
+ * stopping at the first (already-patched) copy left repo-local node_modules
+ * unpatched on global installs (#1190).
  */
-function patchVscodeJsonrpcExports() {
+export function patchVscodeJsonrpcExports(searchRoots = SEARCH_ROOTS) {
   const exportsField = {
     '.': { types: './lib/common/api.d.ts', default: './lib/node/main.js' },
     './node': { node: './lib/node/main.js', types: './lib/node/main.d.ts' },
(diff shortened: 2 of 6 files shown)


tamirdresher (approved):
✅ Reviewed by Squad (Data). Two real bugs, two targeted fixes. The early-return → continue fix for the ESM patcher is obviously correct, and extending REQUIRED_SYNC_HOOKS to include pre-commit/post-commit closes the doctor gap for two-layer state. 6+ regression tests cover exact failure scenarios from #1190.

PR #1464, opened 2026-07-11
fix(cli): follow externalized state in squad copilot and squad rc

omercangumus:
### What
`squad copilot` and `squad rc` now follow the `stateLocation: external` marker: roster reads and writes go to the external state dir instead of the local `.squad/team.md`.

### Why
Closes #1397
Closes #1398
Part of #1402

After `squad externalize`, both commands kept operating on the working-tree copy. `squad copilot` edited (or, on add, effectively scaffolded) a local roster the externalized project never reads — the real team.md in the external dir never learned about @copilot. `squad rc` loaded its agent list from the same dead file, so the remote UI showed an empty or stale roster.

### How
- `commands/copilot.ts`: resolves through `effectiveSquadDir()` — the same external-state-aware resolution already used by export, build, loop, plugin, watch, and doctor (#1443 precedent) — and reads/writes `team.md` in the effective state dir. The `.github/copilot-instructions.md` copy is a working-tree file and is untouched.
- `commands/rc.ts`: the roster load moves into an exported `loadRosterAgents(squadDir)` that routes the `team.md` read through `resolveStateDir()`. Deliberately narrow: the `squadDir` passed to `RemoteBridge` stays the local path, because the bridge derives the copilot session cwd from it by stripping the `/.squad` suffix (`packages/squad-sdk/src/remote/bridge.ts`) — handing it the external state dir would point ACP sessions at the wrong directory. Only the read is redirected.
- Legacy `.ai-team` behavior is unchanged in both commands — no external marker there, resolution falls through to the local path.
- #1399 (`squad cast` agent discovery) is not in this PR; it crosses into squad-sdk's `resolveSquadPaths`/`LocalAgentSource` and lands separately.

### Testing
- `test/cli/copilot.test.ts`: two new tests against a marker-only `.squad/` + external state dir (the layout `squad externalize` leaves behind, same APPDATA/XDG override pattern as the #1443 export tests): add writes the Coding Agent section to the external `team.md` and does not grow a local one; `--off` removes it from the external file.
- `test/cli/rc.test.ts`: two new tests for `loadRosterAgents` — local read without the marker, and the marker redirecting to the external `team.md` while a stale local roster is present (the stale copy must lose).
- `npm run build` — passes. `npm run lint` (tsc) — passes. `npm run lint:eslint` — 0 errors (pre-existing warnings only).
[...]

PR #1443, opened 2026-07-03
fix(cli): resolve externalized state in squad export

omercangumus:
### What
`squad export` now resolves the effective state directory (external when `stateLocation: "external"` is set, local otherwise) instead of reading the local `.squad/` path unconditionally.

### Why
Closes #1396

**Severity: silent data-loss class.** `squad export` is the documented backup/share/migration path. After `squad externalize` (a first-class feature that *moves* `team.md`, `decisions.md`, `routing.md`, `casting/`, `agents/` out of the working tree, leaving only a thin `config.json` marker), export broke in one of two ways:

1. **Misleading hard failure** — the marker-only `.squad/` has no `team.md`, so export dies with `No squad found — run init first`. Following that advice re-scaffolds local state, which sets up:
2. **Silent corruption** — with any local `team.md` present (re-run `init`, partial internalize, manual edits), export reads the local files, reports ✅ success, and ships default scaffolding / stale content instead of the team's real accumulated knowledge. A user who relies on that export as a backup has lost their state without any error.

Anyone using `squad externalize` + `squad export` together is affected, on all platforms.

### Root cause
`export.ts` called `detectSquadDir()` directly and read every state file from `squadInfo.path`. The external-state-aware helper `effectiveSquadDir()` (`packages/squad-cli/src/cli/core/effective-squad-dir.ts`) already exists and is used by `build`, `loop`, `plugin`, `watch`, `doctor`, and shell lifecycle — export was simply never wired through it. (Same class of gap as the sibling issues #1397/#1398/#1399, which are tracked separately under #1402; this PR fixes export only, per one-issue-one-PR.)

### How
- `runExport()` resolves `{ stateDir } = effectiveSquadDir(dest)` and uses `stateDir` for the `team.md` existence check and manifest build.
- `buildManifest()` reads `team.md`, `decisions.md`, `routing.md`, `casting/`, `agents/`, and the `.squad`-local `skills/` source from the effective state dir. Working-tree skill sources (`.copilot/skills`, `.ai-team/skills`) still resolve from `dest` — those live in the repo and are not externalized.
- Non-externalized and legacy `.ai-team` projects are unaffected: without the `stateLocation` marker, `resolveStateDir()` returns the local path unchanged, so behavior is byte-identical to before.

### Repro
```bash
squad init
squad externalize
squad export        # before: "No squad found — run init first" (misleading fatal)
squad init          # follow the bad advice…
squad export        # before: ✅ success — but the export contains default scaffolding, not your team
```

### Regression tests
[...]

Issue #1402, opened 2026-06-26
[External State] Tracking: command and docs integration gaps


dat-intellis:
## Summary

Tracking issue for externalized state (`squad externalize` / `squad internalize`) integration gaps across command behavior and docs consistency.

All linked issues below are part of the same feature area: externalized `.squad` state support.

## Linked child issues

- #1396 — squad export does not resolve externalized state
- #1397 — squad copilot reads and writes local state only
- #1398 — squad rc probes and reads local state directly
- #1399 — squad cast does not follow externalized state for agent discovery
- #1400 — docs and implementation diverge on gitignore behavior
- #1401 — squad internalize copies back without external-dir cleanup

## Source

Findings were verified by direct source inspection and call-chain tracing.

omercangumus:
picking up #1397 and #1398 — both are the same team.md read/write miss, so they'll land as one PR reusing the effective-state resolution from the #1396 export fix. #1399 crosses into squad-sdk agent discovery, that one comes separately.

bradygaster:
The new container hosting guide (#1566) and production ops guide (#1568) both include state-backend selection guidance for containerized deployments, with explicit warnings about branch-based (orphan) state being unsafe for concurrent writes from multiple pods. Both cross-link to this issue as the tracking home for external state gaps. Not a duplicate — these are the deployment docs companion. See umbrella #1564 for full context.

bradygaster:
Closing this integration umbrella as complete. Live state confirms #1396, #1397, #1398, #1399, and #1400 are closed by merged fixes/docs (#1443, #1464, #1465, #1466). The remaining #1401 question is a separate move-vs-copy product decision; current copy-back behavior is explicitly documented by #1466 and #1401 remains open to decide whether it should change.

Issue #871, opened 2026-04-05
cast uses wrong base path for LocalAgentSource in local mode


KenKilty:
## Bug

`cast` reports `No agents found in current session cast.` in local mode even when project agents exist on disk and the interactive shell is already showing them.

## Environment

- `@bradygaster/squad-cli`: `0.9.4-insider.1`
- `@bradygaster/squad-sdk`: `0.9.1`
- `@github/copilot-sdk`: `0.2.1` (overridden locally to work around the known startup/import issue)
- OS: WSL2 Ubuntu / Linux
- Node: `v22.22.1`
- npm: `10.9.4`

## Repro

1. Initialize a local squad repo.
2. Ensure there is at least one project agent directory under `.squad/agents/<name>/charter.md`.
3. Ensure `team.md` has at least one row under `## Members`.
4. Run:

```bash
npm exec squad -- cast
```

## Actual

`cast` prints:

```text
No agents found in current session cast.
```

## Expected

`cast` should list the project agents discovered from `.squad/agents/`.

## Why this looks like a real bug

In the same repo and same install:

- `squad doctor` sees the agents directory and passes health checks
- the interactive shell starts successfully
- the shell header shows a real project agent, e.g. `Lead [IDLE]`

So the project agent files exist and are visible to other parts of the runtime.

## Root Cause

The local `cast` command appears to pass the wrong base path into `LocalAgentSource`.

In the published CLI:

- `dist/cli/commands/cast.js` does:

```js
const paths = resolveSquadPaths(cwd);
const projectSource = new LocalAgentSource(paths.teamDir);
```

But `LocalAgentSource` itself appends `.squad/agents` when resolving local agents:

```js
const AGENT_DIRS = ['.squad/agents', '.ai-team/agents'];
```

In local mode, `resolveSquadPaths(cwd).teamDir` is already the `.squad` directory, so `LocalAgentSource(paths.teamDir)` ends up looking under:

```text
.squad/.squad/agents
```

which is wrong.

## Direct proof

Running this in the affected repo:

```js
const paths = resolveSquadPaths(process.cwd());
console.log(paths.teamDir);

const fromTeamDir = new LocalAgentSource(paths.teamDir);
const fromRepoRoot = new LocalAgentSource(process.cwd());

console.log(await fromTeamDir.listAgents());
console.log(await fromRepoRoot.listAgents());
```

produced:

- `paths.teamDir = /tmp/.../.squad`
- `fromTeamDir.listAgents() = []`
[...]

PR #1465, opened 2026-07-11
fix(cli,sdk): follow externalized state in squad cast agent discovery

### LINKED ISSUE #1399: [External State] squad cast does not follow externalized state for agent discovery
## Summary

`squad cast` does not follow externalized state markers for project agent discovery.

## Evidence

- `packages/squad-cli/src/cli/commands/cast.ts:22,31-36`
  - uses `resolveSquadPaths(cwd)`, then `LocalAgentSource(agentBase)`
- `packages/squad-sdk/src/config/agent-source.ts:36,111-115`
  - `LocalAgentSource` resolves from `<basePath>/.squad/agents`
- `packages/squad-sdk/src/resolution.ts:368-372`
  - `resolveSquadPaths` resolves `teamDir` from `teamRoot` (not external `stateLocation/projectKey`)
- `packages/squad-cli/src/cli/commands/externalize.ts:101-103`
  - externalize sets `teamRoot: '.'` + `stateLocation: 'external'`

## Expected

Agent discovery for cast should resolve through effective state location when externalized.

## Actual

Discovery path is derived from team-root semantics and can remain repo-local.

omercangumus:
### What
`squad cast` now discovers project agents from the external state dir when state is externalized. `LocalAgentSource` accepts an optional explicit agents directory that overrides its `.squad/agents` probing.

### Why
Closes #1399
Part of #1402

`squad externalize` writes `teamRoot: '.'` plus the `stateLocation: external` marker, so an externalized repo lands in `resolveSquadPaths()`'s remote mode with a repo-local `teamDir` — `squad cast` then probes `<repo>/.squad/agents`, which after externalize is a marker dir holding stale leftovers or nothing. The real agents live at `<externalStateDir>/agents` with **no** `.squad` nesting, a directory shape `LocalAgentSource`'s probing can't reach from any base path — which is why this one needs a small SDK addition, unlike the CLI-only #1464.

### How
- `squad-sdk` `config/agent-source.ts`: `LocalAgentSource` gains an optional `agentsDir` constructor parameter that wins over the `.squad/agents`/`.ai-team/agents` probing. Additive — all existing call sites are unchanged, and an override pointing at a missing dir yields an empty list rather than falling back to probing (no silent stale reads). Considered teaching `resolveSquadPaths()` itself about external state instead, but that type is consumed everywhere and changing remote-mode semantics there is a maintainer-level design call; the override keeps the blast radius to one class.
- `squad-cli` `commands/cast.ts`: when `paths.config` carries the external marker, passes `<externalStateDir>/agents` as the override. `resolveExternalStateDir` is imported from the `/resolution` subpath the command already uses — an earlier draft went through the CLI's `effective-squad-dir` helper, which imports the sdk **root barrel**, and that alone added ~5s to the module's cold load in vitest and tripped the cast suite's 5s first-import timeout. Kept the module graph as it was.
- Non-externalized behavior identical: local mode, genuine remote mode (`teamRoot` pointing at another repo), and legacy `.ai-team` all take the unchanged probing path.

### Testing
- `test/config-integration.test.ts` (+2): the override wins over a decoy in the probed `.squad/agents` location; a missing override dir returns empty instead of falling back.
- `test/cli/cast.test.ts` (+1): externalized layout (marker `.squad/` with a stale local agent + real agent in the external dir, APPDATA/XDG override pattern from the #1443 export tests) — cast lists the external agent and not the stale one. Fails without the fix: pre-fix cast finds only the stale local agent.
- All 67 tests across both files pass locally, including the #871 regression tests around `LocalAgentSource` base-path derivation.
- `npm run build` — passes. `npm run lint` (tsc) — passes. `npm run lint:eslint` — 0 errors (pre-existing warnings only).
- Diff hygiene: `git diff` vs `git diff -w` identical, no whitespace-only hunks.

---

### ⚠️ Quick Check
- [x] Changeset added: `.changeset/fix-1399-cast-external-state.md` (minor `@bradygaster/squad-sdk` for the additive public constructor parameter, patch `@bradygaster/squad-cli`)

#### Branch & Commit
- [x] Verified diff contains only intended changes
- [x] PR is not in draft mode

#### Build & Test
- [x] `npm run build` passes
- [x] `npm test` passes (targeted suites 67/67; full-suite state on this Windows box documented on #1464)
- [x] `npm run lint` passes (type check clean)
- [x] `npm run lint:eslint` passes (0 errors)

#### Changeset
- [x] Changeset added (sdk minor + cli patch)

#### Docs
- [x] N/A — behavior fix; the new constructor parameter is documented with TSDoc at the declaration

#### Exports
- [x] N/A — no new module or subpath export; `LocalAgentSource` is already exported from `config/agent-source`

---

### Breaking Changes
None. The new parameter is optional; omitting it preserves the exact previous discovery behavior.

### Waivers
None.

### DIFF
--- a/.changeset/fix-1399-cast-external-state.md
+++ b/.changeset/fix-1399-cast-external-state.md
@@ -1,0 +1,6 @@
+---
+"@bradygaster/squad-sdk": minor
+"@bradygaster/squad-cli": patch
+---
+
+`squad cast` now discovers project agents from the external state dir when state is externalized. `LocalAgentSource` accepts an optional explicit agents directory that overrides the `.squad/agents` probing, since externalized state keeps agents at `<externalStateDir>/agents` with no `.squad` nesting.
--- a/packages/squad-cli/src/cli/commands/cast.ts
+++ b/packages/squad-cli/src/cli/commands/cast.ts
@@ -10,7 +10,7 @@
 import * as path from 'node:path';
 import { LocalAgentSource } from '@bradygaster/squad-sdk/config/agent-source';
 import { resolvePersonalAgents, mergeSessionCast } from '@bradygaster/squad-sdk/agents/personal';
-import { resolveSquadPaths } from '@bradygaster/squad-sdk/resolution';
+import { resolveSquadPaths, resolveExternalStateDir } from '@bradygaster/squad-sdk/resolution';
 import { BOLD, RESET, DIM, GREEN, YELLOW } from '../core/output.js';
 import { fatal } from '../core/errors.js';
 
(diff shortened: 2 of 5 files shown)


tamirdresher (approved):
✅ Reviewed by Squad (Picard). Clean fix — override-not-fallback design is correct (null on missing dir, no silent probing of stale state). Tests cover the exact regression case. Correctly avoids the SDK barrel import perf trap. Minor note: test teardown doesn't clean the stale-agent fixture dir from TEST_ROOT, but not a blocker.

omercangumus:
looked into it — everything that test creates lives under TEST_ROOT (the external dir is TEST_ROOT/global, the stale agent is TEST_ROOT/.squad), and the describe's afterEach already rm -rf's TEST_ROOT after every test. ran the file and checked: zero .test-cast-* dirs left behind. happy to add an explicit per-test cleanup if you'd rather it not lean on the shared teardown

Issue #1211, opened 2026-06-04
State-backend v2: ref-level CAS, recursive deleteDir, stub-name collision, regression coverage


tamirdresher:
# State-backend v2: ref-level CAS, recursive deleteDir, stub-name collision, regression coverage

## Background

This issue tracks the residual concerns from the PR #1200 review that were
intentionally **deferred** out of that PR to keep its scope reviewable. The
in-PR fixes resolved concerns B, G, H in full and concerns C, D in part. The
items below are the remainder.

**Source review:** [PR #1200 review comment 4621356216](https://github.com/bradygaster/squad/pull/1200#issuecomment-4621356216)

**Decision file:** `.squad/workstreams/active/squad-agents-ai/decisions/inbox/picard-pr1200-followup-scope.md`

**Priority:** Ship before promoting `TwoLayerBackend` to the default backend
for fleet / multi-session workloads. The single-user single-session case is
unaffected by concern A and can ship today.

---

### A. Replace bare `update-ref` / `git notes add -f` with CAS-safe forms

**Problem.** The current write path uses:

- `git update-ref refs/heads/<X> <newSha>` — last writer wins, silently
  clobbers concurrent updates from another process.
- `git notes add -f <ref> <sha>` — `-f` is "force overwrite", same
  last-writer-wins behaviour.

Under concurrent writers (`squad watch` + interactive session, two terminals,
fleet) this can drop state without any error.

**Fix.**

1. `update-ref refs/heads/<X> <new> <old>` — compare-and-swap form. Pass
   the SHA read at the start of the write as `<old>`.
2. Replace `git notes add -f` with an optimistic loop: read current note
   blob, build new blob, `git notes add` (no `-f`) and retry on the
   "already exists" / mismatched-parent error.
3. Wrap both in a retry-on-mismatch loop with jittered exponential
   back-off (50 ms → 800 ms, max 5 attempts).
4. On final give-up, throw a typed `StateBackendConcurrencyError` so
   callers can decide how to surface it (toast vs. re-queue vs. abort).

**Acceptance criteria.**

- All write paths in `WorktreeBackend`, `GitNotesBackend`,
  `OrphanBranchBackend`, `TwoLayerBackend` use CAS / optimistic-loop form.
- Grep for `update-ref` and `notes add -f` returns zero hits in
  `packages/squad-sdk/src/`.
- Regression test from concern I below passes (see I).
- Typed `StateBackendConcurrencyError` exported and documented.

---

[...]

PR #1487, opened 2026-07-15
fix(sdk): recursive deleteDir + rename 'external' stub to 'external-stub'

omercangumus:
Part of #1211 — the deleteDir + stub-rename slice (concerns E and F). Concerns A and I already landed with the CAS work (tryUpdateRef, StateBackendConcurrencyError, retry/injector tests), so this is the remainder.

## E — deleteDir left nested keys behind

StateBackendStorageAdapter.deleteDir was a single-level list+delete pass. On git-notes the flat blob has no deletable key for a directory segment, so `deleteDir('agents/x')` deleted nothing under `agents/x/history/2026/` — the nested keys silently survived. deleteDir now deletes each child, then recurses into whatever the backend still lists beneath it: tree-based backends (orphan, two-layer) drop whole subtrees in one commit, git-notes gets the full flat-key walk, and the dual case where a git-notes key is both a leaf and a directory prefix is covered.

WorktreeBackend.delete routes directory keys to a recursive remove so the walk is also safe when the adapter wraps the local backend — this matches the orphan backend, where deleting a tree entry already dropped everything beneath it.

## F — stub rename, adjusted to reality

The issue predicted a config collision with a planned `stateLocation: 'external'`. That feature has since shipped under a *different key*, so the parse collision can't happen as written — but the stub name got worse, not better: `stateBackend: 'external'` now reads like the externalize feature while actually warning and falling back to `'local'`. Renamed the stub to `'external-stub'`, following the existing git-notes → two-layer migration pattern: legacy `'external'` is still accepted everywhere (config.json, `--state-backend`, watch config, `squad init`) and normalized with a one-shot deprecation warning that points at `squad externalize`. `squad init` and watch config write/normalize the canonical name so configs don't trip the warning forever. One stale line in docs/state-backends.md updated to match.

The issue also asked for an upgrade migration that rewrites user configs on disk — the shipped git-notes precedent does read-time normalization instead of disk rewrites, so I followed that. If a disk rewrite is wanted it can ride on migrate-backend later.

## Tests

- new: recursive-delete regression across all 4 backends (the issue's 5-key scenario: write `a/b/c`, `a/b/d`, `a/b/e/f`, `a/b/e/g/h`, `a/other` → `deleteDir('a/b')` → only `a/other` survives), a deleteDirSync variant, the leaf+prefix dual-key case, WorktreeBackend directory delete, external-stub resolution, legacy-alias migration, and one-shot deprecation warning
- test/state-backend.test.ts: 144/144 green; adjacent suites (state-backend-handshake, upgrade-state-backend, upgrade-eperm, test/state/, external-state, cli init, watch-execute, watch-external-loader): all green
- tsc --noEmit clean on squad-sdk

Note on line endings: the touched region of cli-entry.ts is CRLF-committed (the file is mixed-EOL at HEAD), so the two added lines keep CRLF to match their neighbors — same situation as #1464. `git diff -w` matches `git diff` apart from those endings and one real indentation change inside WorktreeBackend.delete.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/fix-1211-deletedir-external-stub.md
+++ b/.changeset/fix-1211-deletedir-external-stub.md
@@ -1,0 +1,6 @@
+---
+"@bradygaster/squad-sdk": patch
+"@bradygaster/squad-cli": patch
+---
+
+Fix StateBackendStorageAdapter.deleteDir leaving nested keys behind: a single-level list+delete pass missed deeper subtrees (e.g. `agents/x/history/2026/log.md` survived a delete of `agents/x`), silently leaking state on git-notes and orphan backends. deleteDir now walks the full subtree on every backend. Also rename the unimplemented `stateBackend: 'external'` placeholder to `'external-stub'` so it can no longer be confused with the real external-state feature (`squad externalize` / `stateLocation: 'external'`); the legacy name is still accepted and normalized with a one-time deprecation warning.
--- a/docs/src/content/docs/features/state-backends.md
+++ b/docs/src/content/docs/features/state-backends.md
@@ -277,7 +277,7 @@ git-native backends (orphan, two-layer) operate in the correct repository contex
 - State backends are **opt-in** — the default is `local` (no behavior change)
 - All backends implement the same interface — agents don't know or care which backend is active
 - Empty directories are automatically pruned after the last file is deleted (orphan backend)
-- The `external` backend type exists as a stub for future external storage (see [External State](./external-state))
+- The `external-stub` backend type is an unimplemented placeholder that falls back to `local` (the legacy name `external` is still accepted with a deprecation warning); for real external storage see [External State](./external-state)
 - State backends are available in the **insider** release channel (`@bradygaster/squad-cli@insider`)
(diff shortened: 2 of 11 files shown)


omercangumus:
CI caught three call sites my local typecheck missed (stale sdk dist locally — CI builds fresh): doctor's configuredStateBackend now maps legacy 'external' to 'external-stub' like its other aliases, and install-hooks/sync treat 'external-stub' as a no-hooks backend alongside 'external'. rebuilt both packages locally and re-ran: doctor 36/36, install-hooks + sync green, cli tsc clean against the fresh sdk build.

tamirdresher (approved):
Approving. Head verified at `e185ae4c25b91d92f0ec7477e839ed936102df62`; CI is green; the 11-file diff cleanly addresses concerns E and F only (not the whole of #1211). Not merging — keeping #1211 open because A/I already landed on `dev` (`tryUpdateRef` CAS + `StateBackendConcurrencyError` + the 5-attempt regression test are present in `packages/squad-sdk/src/state-backend.ts` and `test/state-backend.test.ts`) but the C-residual and D-residual items still need follow-up work.

PR #1459, opened 2026-07-09
docs: add PATH fix instructions to installation troubleshooting

johnazariah:
The troubleshooting section for `squad: command not found` previously told users to check their PATH but didn't explain how to actually fix it when the npm global prefix directory isn't in PATH.

This adds platform-specific instructions for:
- Finding the npm global bin directory with `npm prefix -g`
- Adding it to PATH on macOS/Linux (shell profile)
- Adding it to PATH on Windows (PowerShell environment variable)

Fixes #1458

PR #1488, opened 2026-07-16
docs: show how to fix PATH for custom npm prefixes in installation troubleshooting

### LINKED ISSUE #1458: docs: Troubleshooting section tells users to check PATH but doesn't show how to fix it
## Problem

After running `npm install -g @bradygaster/squad-cli`, the `squad` command is not found on Windows when the npm global prefix is set to a custom directory (e.g. `~/.npm-global`). The current troubleshooting section in the installation docs tells users to *check* their PATH but never explains how to actually *fix* it.

## Steps to reproduce

1. Have a custom npm prefix (e.g. `npm config set prefix ~/.npm-global`)
2. Run `npm install -g @bradygaster/squad-cli` — installs successfully
3. Run `squad` — `squad: The term 'squad' is not recognized`

## Expected

The troubleshooting section should provide the actual fix commands for each platform (adding `npm prefix -g` to PATH), not just diagnostic commands.

## Suggested fix

Update the `squad: command not found` troubleshooting section to include:
- The `npm prefix -g` command to find the actual global bin directory
- Platform-specific instructions for adding it to PATH (bash/zsh profile for macOS/Linux, `[Environment]::SetEnvironmentVariable` for Windows)

omercangumus:
### What
Replaces the diagnostics-only `squad: command not found` entry in `docs/src/content/docs/get-started/installation.md` with the actual fix: `npm prefix -g` to locate the global bin, then per-platform PATH instructions (shell profile export for macOS/Linux, `[Environment]::SetEnvironmentVariable` for Windows). Docs-only — no code change.

### Why
Fixes #1458

The section said "Fix:" but only showed diagnostic commands, so a user with a custom npm prefix could confirm the problem and still not know what to do about it. The old Windows line was also cmd syntax (`echo %PATH% | findstr npm`) that errors in PowerShell, the default shell the docs otherwise assume.

### Testing
- Verified live on Windows 11: `npm install -g --prefix <sandbox> @bradygaster/squad-cli` against the published package — the `squad.cmd` shim lands directly in the prefix root, no `bin\` subdirectory, which is why the doc now spells out the macOS/Linux vs Windows split
- The PowerShell snippet's `[Environment]::` calls syntax-checked with an ephemeral process-scope variable; the real `'User'`-scope write was not run against this machine's PATH, by design
- The bash export line dry-run in a subshell — `$(npm prefix -g)/bin` resolves and prepends as written
- `npm run lint:docs` (markdownlint + cspell) — 0 errors
- `git diff` vs `git diff -w` identical; file stays LF

---

### ⚠️ Quick Check
- [x] No changeset — docs-only change, no governed source/template paths touched

#### Branch & Commit
- [x] Verified diff contains only intended changes (single docs file, +21 −4)
- [x] PR is not in draft mode

#### Build & Test
- [x] `npm run lint:docs` passes
- [x] Build/tests N/A — no code changed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/src/content/docs/get-started/installation.md
+++ b/docs/src/content/docs/get-started/installation.md
@@ -160,16 +160,33 @@ export default defineConfig({
 
 ### `squad: command not found`
 
-Your npm global bin isn't in your PATH. Fix:
+Your npm global bin isn't in your PATH. This usually happens with a custom npm prefix (e.g. `npm config set prefix ~/.npm-global`). Diagnose:
 
 ```bash
 # Check if installed
 npm list -g @bradygaster/squad-cli
 
-# If installed but not found, check PATH:
-echo $PATH | grep npm          # macOS/Linux
-echo %PATH% | findstr npm      # Windows
+# Find where npm puts global commands
+npm prefix -g
 ```
+
+On macOS/Linux, `squad` is installed to `$(npm prefix -g)/bin`. On Windows, it's installed to the prefix directory itself (no `bin` subdirectory). If that directory isn't in your PATH, add it:
+
+**macOS/Linux** — add the bin directory to your shell profile:
+
+```bash
+echo 'export PATH="$(npm prefix -g)/bin:$PATH"' >> ~/.zshrc   # or ~/.bashrc
+source ~/.zshrc
+```
+
+**Windows (PowerShell)** — append the prefix to your user PATH:
+
+```powershell
+$npmPrefix = npm prefix -g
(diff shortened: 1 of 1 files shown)


tamirdresher (approved):

tamirdresher:
Closing in favor of #1459, which now covers the same PATH troubleshooting problem and includes the more complete User PATH handling, duplicate prevention, whitespace trimming, and separate bash/zsh guidance. Thank you for the contribution and review work here.

PR #1492, opened 2026-07-17
fix(cli): externalize/internalize corrupt binary files under .squad/

### LINKED ISSUE #1489: bug: squad externalize corrupts binary files under .squad/ and deletes the originals
## Summary

`squad externalize` moves every file under `.squad/` by reading it as a **utf-8 string** and writing the string back out. Any file that isn't valid UTF-8 text gets its bytes ≥ 0x80 replaced with U+FFFD during the read — and because externalize **deletes the source after the copy**, the corruption is silent and unrecoverable. `squad internalize` copies back through the same lossy path, so the round trip never restores the original.

## Steps to reproduce

Verified on `dev` @ e57b904d (built dist, `squad-cli` 0.11.0), Windows 11, Node v22.21.0. `APPDATA` was overridden so the external dir lands in a sandbox.

```bash
mkdir -p myproj/.squad/assets && cd myproj
echo "# Team" > .squad/team.md
# 1 KB binary file: byte values 0x00-0xFF repeated
node -e "const b=Buffer.alloc(1024); for(let i=0;i<1024;i++) b[i]=i%256; require('fs').writeFileSync('.squad/assets/diagram.png', b)"
sha256sum .squad/assets/diagram.png
# 785b0751fc2c53dc14a4ce3d800e69ef9ce1009eb327ccf458afe09c242c26c9  (1024 bytes)

squad externalize
# ✅ Externalized 2 items. State now at: .../squad/projects/myproj
```

After externalize:

```bash
ls .squad/assets/
# ls: cannot access '.squad/assets/': No such file or directory   ← original gone

sha256sum <external>/projects/myproj/assets/diagram.png
# 36c2611dbed66212ce63c6bb7957c2199c66386fee7f6fafb2f7a3b151bb7a83  (2048 bytes)

dd if=<external>/.../assets/diagram.png bs=1 skip=128 count=12 | od -An -tx1
#  ef bf bd ef bf bd ef bf bd ef bf bd                              ← U+FFFD from byte 0x80 on
```

Every byte ≥ 0x80 became the 3-byte UTF-8 replacement character (1024 → 2048 bytes: 512 low bytes preserved + 512 high bytes × 3). The first 128 bytes survive, which makes the corruption easy to miss at a glance.

## Root cause

Verified on `dev` @ e57b904d:

- `packages/squad-cli/src/cli/commands/externalize.ts:73-76` (file branch) and `:200-215` (`copyDirRecursive`) — both copy via `storage.readSync(src)` → `storage.writeSync(dest, content)`.
- `packages/squad-sdk/src/storage/fs-storage-provider.ts:193` — `readSync` is `readFileSync(path, 'utf-8')`, which decodes invalid sequences to U+FFFD. The string round-trip is lossy for any non-UTF-8 content.
- `externalize.ts:71` / `:78` — the source is deleted (`deleteDirSync` / `deleteSync`) right after the lossy copy, so the only intact copy is destroyed.
- `runInternalize` (`externalize.ts:158-165`) uses the same `copyDirRecursive`, so internalize corrupts in the same way.

## Impact

Core squad state is markdown/JSON today, so text-only projects don't notice. It bites:

- any binary a user or agent keeps under `.squad/` — diagrams or screenshots referenced from decisions.md, PDFs, archives;
- any future binary state: `SqliteStorageProvider` (in-tree, currently unwired) defaults its database to `.squad/squad.db` — the moment that backend becomes reachable, externalize destroys the whole team state db in one command.

Likely also strips POSIX exec bits on scripts (writeSync creates files with default mode) — not verified here, Windows box.

## Suggested fix direction

Copy bytes, not strings: `copyDirRecursive` and the file branch can use the provider's `copy`/`copySync` (`copyFileSync` underneath, already on `FSStorageProvider`) instead of read/write, and only delete the source after the copy is verified. A regression test with a non-UTF-8 fixture would lock it in.

Can pick this up if useful — same area as the deleteDir/stub work in #1487.

Sanjays2402:
Fixes #1489.

`squad externalize` (and `internalize`) copy every file under `.squad/` by reading it as a UTF-8 string and writing the string back out. Any byte >= 0x80 gets decoded to U+FFFD on the way in, so non-UTF-8 files come out corrupted — and since externalize deletes the source right after the copy, the only intact copy is gone. internalize copies back through the same lossy path, so the round trip never restores the original.

Core state is markdown/JSON today so text-only projects don't hit this, but it bites any binary a user or agent keeps under `.squad/` (diagrams/screenshots referenced from decisions, PDFs, archives), and it would destroy the whole state db the moment `SqliteStorageProvider` (`.squad/squad.db`) becomes reachable.

The fix swaps the three `readSync`/`writeSync` copy sites — the externalize file branch, the internalize file branch, and `copyDirRecursive` — for `storage.copySync`, which is `copyFileSync` underneath and already does the recursive mkdir for the destination. Byte-for-byte copy, no string decode.

## Test

Added a regression in `test/external-state.test.ts` that writes a file containing every byte value 0x00–0xFF (invalid UTF-8), runs externalize then internalize, and asserts the externalized copy and the restored copy are both byte-identical to the original. It fails on `main` (`Buffer.equals` is false — the 256-byte file comes back 512 bytes with U+FFFD) and passes with this change. Full `test/external-state.test.ts` suite is green (19 tests), and `tsc --noEmit` on both packages is clean.

### DIFF
--- a/.changeset/fix-externalize-binary-corruption.md
+++ b/.changeset/fix-externalize-binary-corruption.md
@@ -1,0 +1,5 @@
+---
+"@bradygaster/squad-cli": patch
+---
+
+Fix `squad externalize`/`internalize` corrupting non-UTF-8 files under `.squad/`. Both commands copied files through a UTF-8 string round-trip (`readSync`/`writeSync`), which replaced every byte >= 0x80 with U+FFFD and, because the source was deleted right after, destroyed the only intact copy. They now copy bytes via `storage.copySync`, so binary state (diagrams, archives, a future SQLite state db) survives the round trip intact.
--- a/packages/squad-cli/src/cli/commands/externalize.ts
+++ b/packages/squad-cli/src/cli/commands/externalize.ts
@@ -70,11 +70,9 @@ export function runExternalize(projectDir: string, projectKey?: string): void {
       copyDirRecursive(src, dest);
       storage.deleteDirSync?.(src);
     } else {
-      const content = storage.readSync(src);
-      if (content != null) {
-        storage.mkdirSync(path.dirname(dest), { recursive: true });
-        storage.writeSync(dest, content);
-      }
+      // Copy bytes, not a UTF-8 string round-trip — readSync/writeSync
+      // corrupts any non-UTF-8 file, and the source is deleted right after.
+      storage.copySync(src, dest);
       storage.deleteSync?.(src);
     }
     movedCount++;
@@ -158,10 +156,8 @@ export function runInternalize(projectDir: string): void {
     if (storage.isDirectorySync(src)) {
       copyDirRecursive(src, dest);
     } else {
-      const content = storage.readSync(src);
-      if (content != null) {
-        storage.writeSync(dest, content);
-      }
+      // Byte-accurate copy (see externalize branch above).
+      storage.copySync(src, dest);
     }
     movedCount++;
   }
@@ -206,10 +202,8 @@ function copyDirRecursive(src: string, dest: string): void {
     if (storage.isDirectorySync(srcPath)) {
       copyDirRecursive(srcPath, destPath);
     } else {
-      const content = storage.readSync(srcPath);
-      if (content != null) {
-        storage.writeSync(destPath, content);
-      }
+      // Byte-accurate copy (see externalize branch above).
+      storage.copySync(srcPath, destPath);
     }
   }
 }
--- a/test/external-state.test.ts
+++ b/test/external-state.test.ts
@@ -248,6 +248,30 @@ describe('runExternalize / runInternalize', () => {
     expect(readFileSync(path.join(squadDir, 'log', 'session.md'), 'utf-8')).toBe('## Session 1\n');
   });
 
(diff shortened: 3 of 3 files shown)


tamirdresher:
Thanks for the thorough repro and the byte-for-byte fix. Confirmed the corruption on `dev` and confirmed the three `copySync` sites cover all lossy paths. Closes #1489.

Non-blocking follow-up for whoever wires up `SqliteStorageProvider` at `.squad/squad.db`: this exact code path is the reason we must never reintroduce text-mode copying anywhere in `externalize`/`internalize` or `copyDirRecursive` — the regression test in `test/external-state.test.ts` should stay as the guard against that.

tamirdresher (approved):
Externalize binary corruption fix looks correct: text-mode s.readFileSync + trim replaced with copyFileSync (byte-for-byte). Two-pass Buffer-based diff catches the same regression class if it ever comes back. CI (test suite, security review, architectural review, bootstrap protection, diff-size guard, samples build, sdk-exports validation) all green on head 1b667dc2. Approving.

Issue #1556, opened 2026-07-29
squad upgrade regenerates workflows that fail actionlint/shellcheck (SC2086) and downgrade actions/checkout v7 -> v4


serbrech:
## Summary

`squad upgrade` regenerates the `squad-*` workflows from the upstream templates and, in doing so, reverts fixes the downstream repo had already applied. We hit two concrete regressions upgrading a repo from **v0.9.6-build.1 → v0.11.0**.

Repo: `kubefleet-dev/kubefleet` (squad workflows are committed there).

## 1. `squad-heartbeat.yml` fails actionlint + shellcheck (SC2086)

The generated `.github/workflows/squad-heartbeat.yml` writes to `$GITHUB_OUTPUT` unquoted:

```yaml
      - name: Check triage script
        id: check-script
        run: |
          if [ -f ".squad/templates/ralph-triage.js" ]; then
            echo "has_script=true" >> $GITHUB_OUTPUT
          else
            echo "has_script=false" >> $GITHUB_OUTPUT
```

Any repo that runs `actionlint` (with shellcheck available) in CI now fails:

```
.github/workflows/squad-heartbeat.yml:32:9: shellcheck reported issue in this script: SC2086:info:2:29: Double quote to prevent globbing and word splitting [shellcheck]
.github/workflows/squad-heartbeat.yml:32:9: shellcheck reported issue in this script: SC2086:info:4:30: Double quote to prevent globbing and word splitting [shellcheck]
```

Reproduced locally with actionlint **1.7.12** + shellcheck **0.10.0** (same versions as our CI).

**Fix:** quote the redirect target — `>> "$GITHUB_OUTPUT"`.

Worth running `actionlint`/`shellcheck` over the workflow templates in the squad repo's own CI so these don't ship.

## 2. `actions/checkout` downgraded v7 → v4

Our repo had already bumped `actions/checkout` to `v7` in the squad workflows. `squad upgrade` reverted all of them back to `v4` — **12 occurrences across 11 files**:

`squad-ci.yml`, `squad-docs.yml`, `squad-heartbeat.yml`, `squad-insider-release.yml`, `squad-issue-assign.yml`, `squad-label-enforce.yml`, `squad-preview.yml`, `squad-promote.yml` (x2), `squad-release.yml`, `squad-triage.yml`, `sync-squad-labels.yml`

This is a silent dependency downgrade — nothing in the upgrade output flags it, so it is easy to merge without noticing.

[...]

Issue #1502, opened 2026-07-20
[Cross-Squad] Make generated workflow runners and checkout security configurable


desharpe:
Source squad: desharpe/HealthAgent
Request type: cross-repository task delegation
Routing: Booster (CI/CD), with CONTROL/RETRO review
Related: #1493, #1462

## Problem

`@bradygaster/squad-cli` and `@bradygaster/squad-sdk` currently emit GitHub Actions workflows with hardcoded hosted runners and mutable action tags. CLI upgrade also duplicates non-NPM workflow stubs, so repository-local template hardening is not preserved for standalone/global installs. This blocks downstream repositories that require organization-specific self-hosted runner labels and secure checkout defaults.

HealthAgent confirmed that patching installed package files from `postinstall` is not an acceptable fix: it is repository-specific, misses standalone/global CLI usage, and can generate the wrong ecosystem workflow.

## Requested generic fix

- Add shared workflow-generation options supporting scalar or multi-label `runs-on` values without hardcoding a downstream repository's policy.
- Make secure checkout defaults explicit: immutable action references and `persist-credentials: false`.
- Consolidate CLI init/upgrade and non-NPM stubs on the shared generic renderer instead of duplicating workflow YAML generation.
- Expose equivalent runner/security options through the public SDK CI generator.
- Persist optional workflow-generation configuration in `.squad/config.json` so `squad upgrade` can reproduce the intended policy.
- Validate configured runner/action values and reject control characters or mutable action references where custom refs are accepted.
- Keep generated workflows ecosystem-aware for npm, .NET, Go, Python, Java, and unknown projects.

## Acceptance criteria

1. Fresh init and upgrade from a globally installed CLI honor configured scalar and multi-label runners.
2. Public SDK generation produces the same runner and checkout-security behavior as the CLI.
3. Upgrade no longer replaces a configured self-hosted policy with `ubuntu-latest`.
4. Generated checkout steps use immutable pins and disable persisted credentials by default.
5. Generic ecosystem tests cover init, upgrade, standalone/global package behavior, and SDK generation.
6. CLI and SDK receive changesets and the fix is published on npm.

Downstream dependency: desharpe/HealthAgent#190. HealthAgent will remove its rejected local postinstall patch and upgrade to the published fixed packages before its runner-policy PR can merge.

bradygaster:
## Owner & rationale

[...]

PR #1557, opened 2026-07-29
fix(ci): quote GITHUB_OUTPUT/STEP_SUMMARY redirects (SC2086) + add actionlint/shellcheck gate for workflow templates

serbrech:
## What

Downstream repos that run `actionlint` + shellcheck in CI started failing after `squad upgrade`, because Squad's bundled workflow templates redirect to **unquoted** `$GITHUB_OUTPUT` / `$GITHUB_STEP_SUMMARY` in `run:` blocks — shellcheck SC2086 ("double quote to prevent globbing and word splitting").

Since `squad upgrade` regenerates these files wholesale, any downstream fix gets reverted on the next upgrade. So the fix has to land in the templates, and Squad's own CI has to lint them so it cannot ship again.

### SC2086 fixes (unquoted `>> $GITHUB_*` redirects)

Templates (the `squad upgrade` source of truth):
- `.squad-templates/workflows/squad-heartbeat.yml` (canonical source)
- `templates/workflows/squad-heartbeat.yml` (root mirror)
- `packages/squad-cli/templates/workflows/squad-heartbeat.yml`
- `packages/squad-sdk/templates/workflows/squad-heartbeat.yml`

Squad's own workflows (same bug class, found by the new gate):
- `.github/workflows/squad-heartbeat.yml` — 2 occurrences
- `.github/workflows/squad-repo-health.yml` — 10 occurrences across 4 `run:` blocks
- `.github/workflows/squad-ci.yml` — 13 `$GITHUB_STEP_SUMMARY` occurrences across 5 gate steps

Purely additive quoting — no behaviour change.

### New CI gate

`.github/workflows/squad-workflow-lint.yml` — actionlint 1.7.12 + shellcheck 0.10.0, linting **all five** workflow/template directories, including the `.squad-templates/` canonical source and the `templates/` root mirror (not just the `packages/*` mirrors — a canonical-only edit would otherwise ship unlinted, since this job does not run the template sync).

Notes on hardening:
- The actionlint installer is fetched from the **`v1.7.12` tag ref**, not `main`, to avoid a moving-ref RCE surface in CI.
- `ubuntu-latest` ships shellcheck 0.9.0; 0.10.0 is installed explicitly to match the version that surfaced the original report.
- Lints committed state rather than sync-then-lint, so a bug in the canonical source is caught *before* it propagates. Mirror drift is already covered separately by `template-sync.test.ts`.
- Triggered on PR/push with `paths:` filters so it stays off unrelated changes.

## Verification

The proof is this PR's own CI run — actionlint could not be executed locally (no Go/Docker, and `npm install` failed on SSL in the authoring environment), so the pre-merge check is the real validation rather than a local probe.

## What is NOT in this PR

Part 2 of #1556 — `squad upgrade` clobbering `actions/checkout` v7 -> v4 — is deliberately **deferred to a follow-up PR**. It is a more involved change to the upgrade/template-write path (arguably overlapping #1502) and does not belong in a lint fix.

Refs #1556

---
Working as Booster (CI/CD Engineer)

### DIFF
--- a/.changeset/fix-workflow-shellcheck-quoting.md
+++ b/.changeset/fix-workflow-shellcheck-quoting.md
@@ -1,0 +1,21 @@
+---
+"@bradygaster/squad-cli": patch
+"@bradygaster/squad-sdk": patch
+---
+
+Fix shellcheck SC2086 in workflow templates: quote `$GITHUB_OUTPUT` redirects
+
+All `>> $GITHUB_OUTPUT` (and `>> $GITHUB_STEP_SUMMARY`) redirects in `run:` blocks were unquoted, causing `actionlint` + shellcheck to report SC2086 (double quote to prevent globbing and word splitting) in downstream repos that run `actionlint` in their CI. The fix is purely additive quotes around the variable; behaviour is unchanged.
+
+**Files fixed:**
+- `.squad-templates/workflows/squad-heartbeat.yml` (canonical source — synced to all mirrors)
+- `templates/workflows/squad-heartbeat.yml`
+- `packages/squad-cli/templates/workflows/squad-heartbeat.yml`
+- `packages/squad-sdk/templates/workflows/squad-heartbeat.yml`
+
+**Squad's own workflows also fixed:**
+- `.github/workflows/squad-heartbeat.yml`
+- `.github/workflows/squad-repo-health.yml`
+- `.github/workflows/squad-ci.yml`
+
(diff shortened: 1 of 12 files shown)


tamirdresher (approved):
Verified the diff: all quoting fixes are correct and consistent across all 5 locations (canonical + 4 template mirrors), no missed unquoted GITHUB_OUTPUT/STEP_SUMMARY patterns repo-wide. New squad-workflow-lint.yml gate uses least-privilege permissions, properly pinned action versions (verified SHA matches v7 tag), and hardened installer scripts. All CI checks green including the new lint job. Minor non-blocking notes: one stale Copilot review thread can be marked resolved, and a small unrelated dead-variable cleanup in size-regression-report.yml is harmless scope creep. Approving.

PR #1558, opened 2026-07-29
chore(squad): merge decision inbox and summarize oversized agent histories

serbrech:
## What

Scribe maintenance pass produced during the issue #1556 work. Split out of #1557 so that PR stays scoped to the shellcheck/actionlint gate.

**No product code, no workflows, no templates.** `.squad/` team state only.

### Decision inbox merged (6 entries → `decisions.md`, inbox cleared)

- `booster-ci-deletion-guard.md`
- `retro-copilot-git-safety.md`
- `flight-versioning-policy.md`
- `booster-release-skill-v094.md`
- `procedures-fix-coordinator-inline-dispatch-gate.md`
- `booster-workflow-template-linting.md` (new, from this session)

`decisions.md`: 12,151 → 16,030 bytes — still under the 20KB archival threshold, so no entries were archived.

### Agent histories summarized (over the 15KB threshold)

| Agent | Before | After |
|-------|--------|-------|
| eecom | 33,933 | 5,770 |
| pao | 30,999 | 5,556 |
| _alumni/kobayashi | 26,305 | 3,489 |
| flight | 23,236 | 5,661 |
| fido | 22,844 | 6,007 |
| procedures | 18,945 | 4,082 |

This is where the ~1,700 deleted lines come from — it is compaction of team memory, not lost work. Prior content remains in git history.

### Cross-agent updates

Short team-update notes appended to `fido/history.md` (new CI quality gate exists) and `surgeon/history.md` (a changeset landed; new lint job affects release gating).

## Reviewer note

Worth a skim of the summarized histories to confirm nothing load-bearing was compressed away — summarization is lossy by design, and these six files carried a lot of accumulated context.

Refs #1556

---

## ⛔ Blocked on #1559 (merge that first)

`Security Review` fails here on a **false positive** — the scanner flags `.squad/decisions.md` prose that *documents* the `git add .` prohibition as if it were executing it.

The fix is #1559. It cannot be included in this PR: `squad-repo-health.yml` uses `pull_request_target` and checks out the **base branch** for trusted scripts, so CI always runs the `dev` copy of `security-review.mjs`. The fix only takes effect once it is on `dev`.

**Merge order: #1559 → then re-run checks here.**

PR #1559, opened 2026-07-29
fix(security): stop unsafe-git scanner flagging prose in append-only squad logs

serbrech:
## Problem

`scripts/security-review.mjs` check #4 (`unsafe-git`) iterates over **every** changed file with no file-type filter — unlike the other checks, which scope to `workflowFiles` or `jstsFiles`. Any markdown file containing the literal string `git add .` on an added line is reported as an **error**.

This fires on prose that *documents the prohibition*. Concretely, it blocked #1558 by flagging Squad's own decision log:

```json
{
  "category": "unsafe-git",
  "severity": "error",
  "message": "Unsafe git operation: `git add .` ...",
  "file": ".squad/decisions.md",
  "line": 257
}
```

…where line ~262 reads:

> **What:** Added mandatory Git Safety section to copilot-instructions.md: prohibits `git add .`, requires feature branches and PRs …

The scanner flagged the very rule it was documenting.

**Why it surfaced now:** `.squad/decisions/inbox/` is gitignored, so inbox entries never appeared in PR diffs. Once Scribe merged them into `.squad/decisions.md`, those lines became `+` lines for the first time.

## Fix

Adds `UNSAFE_GIT_EXCLUDED_PATHS` and skips the `unsafe-git` check for **append-only memory/log files** — records of what happened, never executed:

- `.squad/decisions.md`
- `.squad/agents/*/history.md`
- `.squad/log/**`, `.squad/orchestration-log/**` (gitignored; listed defensively)

**Deliberately still in scope — instruction surfaces:**

- `.github/copilot-instructions.md`
- `.squad/agents/*/charter.md`, `.squad/team.md`, `.squad/routing.md`, `squad.agent.md`
- `.squad/templates/**`

A malicious PR could swap a prohibition for an instruction in those files, so they must stay scannable. The line is drawn at files that are *structurally* append-only versus files agents read as active instructions.

## Verification

Detection still works — a probe file with `git add .` and `git push --force` was correctly caught:

```
"file": "scripts/__scan-probe.sh", "line": 3  → git add .
"file": "scripts/__scan-probe.sh", "line": 4  → git push --force
🔒 Security review: 2 error(s).
```

Probe was deleted; no trace remains. Running the patched scanner against this PR's own diff returns `✅ No security concerns found.`

## Why this is a separate PR

`squad-repo-health.yml` uses `pull_request_target` and **checks out the base branch** for trusted scripts (correct — it prevents fork PRs from executing arbitrary code with a write token). That means CI always runs the `dev` copy of `security-review.mjs`, so this fix cannot take effect on the PR that contains it. It has to land on `dev` first.

**Merge order:** this PR → then #1558 goes green.

## Known limitation (not fixed here)

The pattern `/git\s+add\s+\./` also matches *scoped* paths like `git add .squad/`, which is the safe, recommended form. That is a pattern-precision bug distinct from this scoping bug — filed as follow-up rather than bundled in.

Refs #1556

### DIFF
--- a/scripts/security-review.mjs
+++ b/scripts/security-review.mjs
@@ -174,6 +174,21 @@ for (const file of jstsFiles) {
 }
 
 // 4. Unsafe git operations
+//
+// Excluded paths: append-only memory/log files that document history and prohibitions.
+// These are never executed — they record what happened, not what to do.
+// Instruction surfaces (.github/copilot-instructions.md, agent charters, squad.agent.md,
+// team.md, routing.md) are deliberately kept in scope: a malicious PR could replace a
+// prohibition with an instruction, and we want the scanner to catch that.
+// .squad/log/ and .squad/orchestration-log/ are gitignored (never in PR diffs) but
+// listed here for documentation completeness.
+const UNSAFE_GIT_EXCLUDED_PATHS = [
+  /^\.squad\/decisions\.md$/,
+  /^\.squad\/agents\/.+\/history\.md$/,
+  /^\.squad\/log\//,
+  /^\.squad\/orchestration-log\//,
+];
+
 const GIT_UNSAFE_PATTERNS = [
   { pattern: /git\s+add\s+\./, label: 'git add .' },
   { pattern: /git\s+add\s+-A/, label: 'git add -A' },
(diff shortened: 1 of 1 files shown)


tamirdresher (approved):
Reviewed the exclusion logic in scripts/security-review.mjs. The unsafe-git false-positive fix is correct and the nested-path regex issue flagged by the automated review is already resolved in this commit. One trade-off worth noting for the record: .squad/decisions.md and agent history.md files are loaded into every agent's context on spawn, so excluding them from the unsafe-git scanner does create a theoretical blind spot if a maliciously-crafted 'decision' entry contained an actual unsafe git instruction disguised as prose. Accepting this trade-off for now since these files are structurally append-only and reviewed via normal PR process before landing, and the false-positive cost of NOT excluding them was actively blocking legitimate PRs. Approving with this caveat on record; worth revisiting if the decision-inbox pipeline becomes more automated in the future.

PR #1587, opened 2026-08-04
feat(gh-aw): add Squad shared component for GitHub Agentic Workflows

bradygaster:
## What this does

Adapts Peli de Halleux's proven **Squad × gh-aw** (GitHub Agentic Workflows) integration from [`github/gh-aw`](https://github.com/github/gh-aw/blob/main/.github/workflows/shared/squad.md) into this repo. Peli built and demonstrated it and explicitly offered it for upstreaming — **this is an adaptation of his design, not a rewrite.** Credit to Peli.

There is **no issue to close** — this was dispatched directly by Brady off Peli's demonstration.

## The architecture (Peli's — preserved exactly)

1. **Activation job** (`jobs.activation.pre-steps`) runs `npx @bradygaster/squad-cli init`, then uploads `.squad/` + `.github/agents/squad.agent.md` as a `squad-state` artifact.
2. **Agent job** (`steps:`) downloads and restores that artifact. **The Squad CLI never runs in the agent job.**

This sidesteps gh-aw's network firewall, which only constrains the agent job — activation has open egress. The App-token → `SQUAD_GITHUB_TOKEN` → `github.token` auth ladder is kept intact.

## Deliverables

- **`.github/workflows/shared/squad.md`** — the importable shared component (no `on:` field).
- **`.github/workflows/squad-backlog-triage.md`** (+ compiled `.lock.yml`) — sample `workflow_dispatch` workflow that imports the component and triages this repo's own issues.
- **`docs/src/content/docs/features/gh-aw.md`** (+ nav entry) — feature documentation.

## Four intentional changes from Peli's reference

1. **Import shapes documented** — repo-local (`shared/squad.md`) and SHA-pinned remote (`bradygaster/squad/.github/workflows/shared/squad.md@<sha>`), verified against gh-aw's [imports reference](https://github.github.com/gh-aw/reference/imports/).
2. **`--state-backend local` pinned** on the init command — the compiled agent invocation passes `--disable-builtin-mcps`, so Squad's `state-mcp` bridge can't load; a non-local backend would fail silently.
3. **Dropped the inaccurate `--agent squad` claim** from Peli's header comment — replaced with an accurate description of gh-aw's `.github/agents/*.agent.md` sub-agent pickup (there is no `--agent` flag on the compiled Copilot invocation).
4. **`SQUAD_CLI_VERSION` default kept at `0.11.0`** with a "bump every release" note. See follow-up below.

## Validation

Compiled locally with `gh aw compile` (**gh-aw v0.81.6, strict mode, 0 errors**). The compiled lock confirmed:

- `npx … init --preset default --state-backend local` (line 122)
- `GH_AW_SUB_AGENT_DIR=".github/agents"` / `GH_AW_SUB_AGENT_EXT=".agent.md"` pickup (lines 519–520)
- `--disable-builtin-mcps` on the agent invocation (line 814)
- **no `--agent` flag** anywhere

`test/docs-build.test.ts` passes 22/22; cspell + markdownlint clean. Action versions (`create-github-app-token@v3.2.0`, `upload-artifact@v7.0.1`, `download-artifact@v8.0.1`) match Peli's reference exactly.

> The committed `.lock.yml` was compiled with gh-aw v0.81.6. If the repo standardizes on a different gh-aw version, recompile with `gh aw compile`.

## Known limitations (documented honestly, in the feature doc)

- **Cross-run state does not persist** — the `squad-state` artifact is a within-run handoff (`retention-days: 1`). Casting registry, session logs, and Scribe output don't survive between runs. Solving this (artifacts / Actions cache / dedicated branch) is unsolved and out of scope.
- **npm required in the activation job** — works on GitHub-hosted runners (open egress); a self-hosted runner behind a corporate npm-blocking firewall would fail. Out of scope.
- **No MCP state bridge in the agent job** (`--disable-builtin-mcps`).

## Drive-by fix

`EXPECTED_GUIDES` in `test/docs-build.test.ts` was missing `agent-framework-integration` (added in #1574, never synced) — the docs-build test was already failing on `dev`. Fixed here since EXPECTED_* sync is FIDO's charter responsibility and the test file was already being edited.

## Scope / safety

`repo-health`-shaped: only `.github/`, `docs/`, `test/`, and `cspell.json`. **No `packages/*/src/` changes → no changeset required.** Reference copies under `.squad/proposals/reference/` were not touched.

## Security note (from `gh aw compile`)

The compile flagged two new restricted secrets — `SQUAD_GITHUB_APP_PRIVATE_KEY` and `SQUAD_GITHUB_TOKEN`. Both come from the shared component's optional auth ladder (mint a GitHub App token / fall back to a PAT for cross-org `squad init`), are referenced only in the activation job, and default to unused (`github.token`) when unset. No credential exfiltration or unexpected egress introduced.

⚠️ Needs squad member review before merging.

Design credit: **Peli de Halleux** — https://github.com/github/gh-aw/blob/main/.github/workflows/shared/squad.md


bradygaster:
Verification note — the `SQUAD_CLI_VERSION` pin in this component has a known caveat that is **not** a defect in this PR.

Published `@bradygaster/squad-cli@0.11.0` declares `"@bradygaster/squad-sdk": "*"`, so pinning the CLI does not pin the SDK — and the SDK carries the presets and agent charters. Today `*` resolves to `0.11.0` by coincidence; the first `0.12.0` SDK publish will silently change what this workflow runs.

Tracked separately in #1588. Out of scope here (this PR touches no `packages/*/src/`), but it affects how much the version pin in `shared/squad.md` is actually worth.

Everything else verified against the published tarballs rather than local source, since that is what the workflow consumes:
- `--state-backend` and `--preset` both present in published `squad-cli@0.11.0`
- published `squad-sdk@0.11.0` ships `dist/presets/builtin/default/` with 5 charters
- `gh aw compile` v0.81.6 strict → 0 errors

Still unproven: no end-to-end gh-aw run has ever executed.

bradygaster:
Applied Peli's review suggestion: added `--ignore-scripts` to the activation-job `npx` (commit 20d6681).

**What changed**
- `.github/workflows/shared/squad.md` — flag added before the package spec (`npx --yes --ignore-scripts "@bradygaster/squad-cli@…"`), plus a header comment explaining why it's safe so nobody removes it later. Credit to @pelikhan for the suggestion.
- `.github/workflows/squad-backlog-triage.lock.yml` — recompiled with `gh aw compile` (v0.81.6, strict, 0 errors). The lock's run line now matches the source byte-for-byte.
- `docs/src/content/docs/features/gh-aw.md` — activation-job row updated so docs and workflow agree.

**Why it's safe (not a blind accept)**
`@bradygaster/squad-cli@0.11.0` *does* ship a postinstall (`patch-esm-imports.mjs` + `patch-ink-rendering.mjs`, both in the tarball). `patch-esm-imports` is load-bearing — it patches `vscode-jsonrpc` so Node 22+ strict ESM resolves subpath imports for Copilot SDK sessions (#449). **But `squad init` is pure file scaffolding and never opens an SDK session.**

A/B verified on Node v24.16.0 (the version the patch targets):

| Test | Command | Result |
|---|---|---|
| A | `npx --yes --ignore-scripts …@0.11.0 init --preset default --state-backend local` | exit 0, 5 agents, 100 md files |
| B | same without `--ignore-scripts` | exit 0 |

`diff -r` of the two `.squad/` trees: only differences are the `.first-run`/`updated_at` timestamps and the temp-dir name. Agent lists identical; `.github/agents/squad.agent.md` **byte-identical**. So `--ignore-scripts` skips no real work for `init` while blocking lifecycle-script execution from any compromised transitive dep — exactly what security scanners look for.

**Validation:** `docs-build.test.ts` 22/22, cspell clean on in-scope docs. Landed on `squad/gh-aw-shared-component` only.

(Re: the 109KB compiled lock shrinking in the next AW release — noted, informational, no action here.)

.github/workflows/shared/squad.md
```diff
@@ -0,0 +1,138 @@
...
+#   wiring this into release tooling automatically.
+#
+# State backend is pinned to `local` on purpose: the compiled agent invocation passes
+# `--disable-builtin-mcps`, so Squad's `state-mcp` bridge does not load in the agent job.
+# A non-local backend would fail silently, so `squad init` is run with `--state-backend
+# local`. Cross-run state does NOT persist (the squad-state artifact is a within-run
+# handoff with retention-days: 1) — every run starts from a fresh `squad init`.
+engine:
+  id: copilot
+
```
pelikhan, line 65: @copilot add `agent: squad` to emit --agent squad

tamirdresher:
Reviewed. Two things — one safety note, one suggestion.

## 1. `--state-backend local` is protecting more than the PR credits it for

The PR justifies the pin as *"the agent job runs with `--disable-builtin-mcps`, so the state-mcp bridge can't load."* True — but it's also gating this, in `packages/squad-cli/src/cli/core/init.ts`:

```ts
// Lift those mutable files onto the squad-state orphan branch
// and remove the working-tree copies so the backend is the
// single source of truth post-init.
const lifted = liftInitMutableStateOntoOrphan(dest);
```

That strips `decisions.md` and every agent's `history.md` out of the working tree. It's guarded by `if (options.stateBackend === 'orphan' || options.stateBackend === 'two-layer')` — so on a repo with committed state, the `local` pin is the only thing stopping init from deleting it from the checkout before the agent ever sees it.

Not a bug in this PR — it's correct as written. But the reason is undocumented, so whoever fixes the MCP issue later could reasonably drop the pin as no-longer-needed and silently reintroduce it. Worth a line in the header comment.

## 2. Could we detect an existing squad and skip the bootstrap entirely?

On a repo that already commits `.squad/`, `init` is close to a no-op — `skipExisting: true`, templates are `overwriteOnUpgrade: false`, and `squad.agent.md` only gets `stampVersion()`. Everything else prints `already exists — skipping`.

So today, for those repos, we install from npm, resolve the SDK unpinned (#1588), skip everything, and upload an artifact duplicating files already in git.

`detectSquadDir()` already exists in `src/cli/core/detect-squad-dir.ts`, but the workflow only needs a file test:

```yaml
- name: Detect existing Squad
  id: squad-detect
  run: |
    if [ -f .squad/team.md ] && [ -f .github/agents/squad.agent.md ]; then
      echo "present=true" >> "$GITHUB_OUTPUT"
    else
      echo "present=false" >> "$GITHUB_OUTPUT"
    fi

# then add to the three existing activation steps:
#   Mint token:  if: steps.squad-detect.outputs.present != 'true' && vars.SQUAD_GITHUB_APP_ID != ''
#   Init:        if: steps.squad-detect.outputs.present != 'true'
#   Upload:      if: success() && steps.squad-detect.outputs.present != 'true'
```

The agent job needs no changes — its restore step is already `continue-on-error: true`, so with no artifact it quietly no-ops and the checked-out files win.

That takes npm off this path entirely for committed-squad repos, which also takes #1588 off it. Probably wants a `vars.SQUAD_FORCE_INIT` escape hatch for repos that would rather float on the latest CLI, and a note that the committed version becomes authoritative — which I think is the right default, just worth stating.

Also worth flagging separately: this is independent of `ambient-folders`. That optimizes *how* the bootstrap ships files; detection removes the need to ship them at all for the committed case.

tamirdresher:
Follow-up to my previous comment — here's the concrete patch for the detection idea, so it's directly usable rather than just a question.

## 1. `.github/workflows/shared/squad.md` — activation job

```diff
 jobs:
   activation:
     pre-steps:
+      - name: Detect committed Squad
+        id: squad-detect
+        env:
+          SQUAD_FORCE_INIT: ${{ vars.SQUAD_FORCE_INIT }}
+        run: |
+          set -euo pipefail
+          if [ "${SQUAD_FORCE_INIT:-}" = "true" ]; then
+            echo "present=false" >> "$GITHUB_OUTPUT"
+            echo "SQUAD_FORCE_INIT is set — bootstrapping from npm."
+          elif [ -f .squad/team.md ] && [ -f .github/agents/squad.agent.md ]; then
+            echo "present=true" >> "$GITHUB_OUTPUT"
+            echo "Committed Squad detected — skipping npm bootstrap."
+          else
+            echo "present=false" >> "$GITHUB_OUTPUT"
+            echo "No committed Squad — bootstrapping from npm."
+          fi
+
       - name: Mint Squad GitHub App token
         id: squad-app-token
-        if: ${{ vars.SQUAD_GITHUB_APP_ID != '' }}
+        if: ${{ steps.squad-detect.outputs.present != 'true' && vars.SQUAD_GITHUB_APP_ID != '' }}
         uses: actions/create-github-app-token@v3.2.0
         ...

       - name: Initialize Squad team
+        if: ${{ steps.squad-detect.outputs.present != 'true' }}
         env:
           SQUAD_CLI_VERSION: ${{ vars.SQUAD_CLI_VERSION }}
         ...

       - name: Upload Squad state artifact
-        if: success()
+        if: ${{ success() && steps.squad-detect.outputs.present != 'true' }}
         uses: actions/upload-artifact@v7.0.1
         ...
```

**The agent job needs no changes.** Its restore step is already `continue-on-error: true`, so when no artifact was uploaded it no-ops and the checked-out files are used as-is.

`SQUAD_FORCE_INIT` is read via `env:` rather than inlined into the script, to keep the repo variable off the shell command line.

## 2. Header comment — two additions

```
# Detection: if the consuming repo already commits `.squad/` and
# `.github/agents/squad.agent.md`, the activation job skips the npm install,
# `squad init`, and the artifact upload entirely — the agent job's own
# checkout already has them. Set `vars.SQUAD_FORCE_INIT=true` to bootstrap
# from npm anyway (e.g. to float on the latest CLI rather than the committed
# version). Note this makes the committed Squad version authoritative for
# these repos; upgrade deliberately with `squad upgrade` and commit the diff.
#
# `--state-backend local` is load-bearing for TWO reasons, not one:
#   1. the agent runs with `--disable-builtin-mcps`, so the state-mcp bridge
#      cannot load and a non-local backend would fail silently; and
#   2. `init` calls `liftInitMutableStateOntoOrphan()` for the `orphan` and
#      `two-layer` backends, which REMOVES `decisions.md` and every
#      `agents/<n>/history.md` from the working tree. On a repo with committed
#      state that would strip it out of the checkout before the agent runs.
# Do not drop this pin when (1) is eventually fixed.
```

## 3. Prompt text — two lines currently assume `init` always ran

```diff
-Squad's team state (`.squad/`) and its Copilot custom agent
-(`.github/agents/squad.agent.md`) were already initialized during activation and
-restored into this checkout before you started — do **not** install Squad or run
-`squad init` yourself, and do **not** try to reach the npm registry.
+Squad's team state (`.squad/`) and its Copilot custom agent
+(`.github/agents/squad.agent.md`) are already present in this checkout — either
+committed to the repository, or generated during activation — do **not** install
+Squad or run `squad init` yourself, and do **not** try to reach the npm registry.
```

```diff
-- State does **not** carry over between runs. The casting registry, session
-  logs, and any Scribe output you produce live only for this run. Do not assume
-  a prior run's decisions are still on disk — start from what `squad init`
-  produced plus the repository's own tracked history (issues, PRs, `.squad/`
-  files committed to the repo).
+- Anything you write to `.squad/` during this run is discarded when the run ends.
+  The casting registry and session logs do not carry over. Base your work on what
+  is on disk now (the committed `.squad/` files, or what `squad init` produced)
+  plus the repository's tracked history — do not assume a prior run's decisions
+  survived.
```

---

Net effect for a repo that commits `.squad/`: no npm, no SDK resolve at run time (so #1588 stops applying to this path), no artifact round-trip, and no ambiguity in the prompt about which team is in play. For repos that don't commit it, behaviour is unchanged.

bradygaster:
Closing — superseded by the new unified /squad workflow design (see epic #1612). The shared component will be rebuilt as #1613.

PR #1660, opened 2026-08-08
fix(cli): route watch/loop capability state through the effective state dir

### LINKED ISSUE #1490: bug: squad watch capabilities read and write the local .squad/ on externalized projects — silent no-ops
## Summary

`squad watch` resolves externalized state for its **startup reads only**. `watch/index.ts` calls `effectiveSquadDir()` and uses the external-aware `stateDir` to find `team.md` / `routing.md` — but then hands every capability `teamRoot = path.dirname(local .squad)`, and the capabilities all build `path.join(context.teamRoot, '.squad', ...)`. On an externalized project the local `.squad/` contains only `config.json`, so watch starts cleanly and then every state-reading capability silently no-ops against an empty directory while the real state sits in the external dir.

This is the watch-shaped remainder of the #1402 family: export (#1396), copilot (#1397), rc (#1398), and cast (#1399) got fixed; #949 fixed watch's startup reads but the capability layer was never in scope.

## Steps to reproduce

Verified on `dev` @ e57b904d (built dist, `squad-cli` 0.11.0), Windows 11, Node v22.21.0, sandboxed `APPDATA`.

```bash
mkdir -p watchproj/.squad/decisions/inbox && cd watchproj
echo "# Team" > .squad/team.md
echo "# Routing" > .squad/routing.md
for i in 1 2 3 4 5 6; do echo "# decision $i" > ".squad/decisions/inbox/d$i.md"; done

squad externalize
# ✅ Externalized 3 items.

ls .squad/                                          # config.json          ← local is empty
ls <external>/projects/watchproj/decisions/inbox/   # d1.md ... d6.md      ← 6 files, over the >5 threshold
```

Then resolve the watch context exactly as `watch/index.ts:678-681` does and run the decision-hygiene preflight (compiled dist modules, no mocks):

```
stateDir (startup reads, external-aware): ...\appdata\squad\projects\watchproj
teamRoot handed to capabilities        : ...\watchproj
decision-hygiene preflight: {"ok":false,"reason":"no decision inbox directory found"}
```

Six inbox files over the merge threshold, and the capability disables itself. Before externalize the same preflight passes.

## Root cause

`packages/squad-cli/src/cli/commands/watch/index.ts:678-681` — `stateDir` is used for `team.md`/`routing.md` only; `teamRoot` is derived from the **local** `.squad/` and passed into `WatchContext`. Every consumer joins `.squad/` from it:

- `capabilities/decision-hygiene.ts:20,29` — inbox scan (repro above)
- `capabilities/cleanup.ts:77,93`
- `capabilities/notes-promote.ts:66,91`
- `capabilities/retro.ts:29` — `.squad/log`
- `capabilities/execute.ts:99` — `.squad/ralph-instructions.md` (the #1314 escape hatch silently stops working after externalize)
- `external-loader.ts:36` — `.squad/capabilities/`, so custom capabilities (#918) stop loading too

`watch/config.ts:71` (config.json) is correct as-is — config.json is deliberately KEEP_LOCAL. `pid-tracker.ts:22` (`.watch-pids`) is arguably fine local since it's machine-scoped.

Beyond the silent no-ops there's a split-brain risk: capabilities that trigger writes (decision-hygiene and notes-promote spawn an agent with `cwd = teamRoot` and prompts that reference `.squad/decisions...`) would write into the local `.squad/` of an externalized project, and a later `squad internalize` copies the external files over the local ones. Read-side no-op is what's verified above; the write side is the same paths by construction.

## Suggested fix direction

Add the effective state dir to `WatchContext` (e.g. `context.stateRoot`, set from `effectiveSquadDir().stateDir`) and have capabilities join state paths from it, keeping `teamRoot` for repo-scoped operations (git, gh, spawn cwd). Same shape as the copilot/rc fix in #1464.

Can pick this up if useful.

omercangumus:
### What
`squad watch` and `squad loop` now thread the effective state directory (`stateRoot`) through `WatchContext` and every capability, instead of each capability rebuilding `.squad/` paths from `teamRoot` on its own.

### Why
Closes #1490

Full repro and root cause are in the issue — the short version: `runWatch`/`runLoop` already call `effectiveSquadDir()` and use the external-aware `stateDir` for `team.md`/`routing.md`, but every capability was still handed plain `teamRoot` and joined `.squad/` onto it directly. After `squad externalize`, the local `.squad/` is a marker-only stub (`config.json` plus a few other `KEEP_LOCAL` files), so every capability doing `path.join(teamRoot, '.squad', ...)` silently found nothing: decision-hygiene's inbox count, cleanup's scratch/log/orchestration-log pruning, retro's staleness check, subsquad discovery, the `capabilities/` external loader, and the `ralph-instructions.md` lookup in `execute.ts`.

### How
- `types.ts`: added `stateRoot: string` to `WatchContext`.
- `watch/index.ts` / `loop.ts`: set `stateRoot: stateDir` in `baseContext` — `stateDir` was already computed via `effectiveSquadDir()` in both files, just never threaded further. The compiler caught `loop.ts`'s missing field the moment `stateRoot` became required — same bug, second entry point, would've been an easy miss by hand.
- `decision-hygiene.ts`, `cleanup.ts`, `retro.ts`: swapped `path.join(context.teamRoot, '.squad', ...)` for `path.join(context.stateRoot, ...)`.
- `discoverSubSquads()` (private in `watch/index.ts`, only caller was `runWatch` itself): same swap, now takes `stateRoot` directly. Exported it (matching the existing `reportBoard` precedent in the same file) so it's actually testable.
- `loadExternalCapabilities()` and `buildAgentPrompt()`: both took a bare `teamRoot` and joined `.squad/` internally, with existing test suites built around that exact contract (7 tests for the loader, 4 for the prompt builder). Rather than changing what they join, both got a new optional param (`capabilitiesDir` / `stateDir`) that wins when passed — zero behavior change for any existing 2-arg caller, callers with the full context now pass `context.stateRoot` as the extra arg.
- `notes-promote.ts`: **left unchanged**, differs from what the issue originally listed. It only reads `.squad/config.json` for backend-type detection — `config.json` is deliberately `KEEP_LOCAL` in `externalize.ts`, and the two-layer backend's actual state (git-notes + orphan branch) lives in git refs, not the filesystem, so externalization doesn't touch any of it. Flagging this as a correction to the issue's own root-cause list, not silently dropping scope.

### Testing
Regression tests added to the existing suites, all following each file's established mock/fixture style — no new conventions:
- `test/cli/watch-capabilities.test.ts`: new `externalized state (#1490)` blocks under `CleanupCapability` and `DecisionHygieneCapability`, a new `RetroCapability` describe block (didn't exist before), and a `buildAgentPrompt` override test — each gives `teamRoot`/`stateRoot` different fake paths and asserts the mocked storage calls only ever touch `stateRoot`.
- `test/cli/watch-external-loader.test.ts`: new block proving `capabilitiesDir` wins over the `teamRoot`-derived default, and that omitting it preserves the old default exactly (existing 7 tests untouched).
- `test/watch-subsquad-discovery.test.ts` (new file): real-filesystem test for the now-exported `discoverSubSquads`, including one that writes a subsquad only under an "external" dir and shows it's invisible from the local `.squad/` — the actual shape of the bug.
- Ran every new/changed test against unfixed `dev` first (stashed the 8 source files, kept the tests): 12 failures — the exact ones exercising the fix (Windows-path assertions, missing export, override ignored). Restored the source, all pass.

```
npx vitest run test/cli/watch-capabilities.test.ts test/watch-cleanup.test.ts test/watch-notes-promote.test.ts test/cli/watch-agent-spawn.test.ts test/cli/watch-external-loader.test.ts test/watch-subsquad-discovery.test.ts
 Test Files  1 failed | 5 passed (6)
      Tests  1 failed | 97 passed (98)
```
The 1 failure (`buildAgentPrompt > checks .squad/ralph-instructions.md inside teamRoot`) is a pre-existing Windows path-separator mismatch, confirmed present on unmodified `dev` — untouched by this PR, same function, different (2-arg, unchanged-behavior) call path.

Full-suite (`npx vitest run`, fresh `dist` for both packages): 87 failures / 7232 passed. Checked every uniquely-named failing file against unmodified `dev`: `test/cli/watch.test.ts` (cold dynamic-import 5s timeout) and `test/cli/upgrade.test.ts` (`casting_policy_version` 1.1 vs 1.2, unrelated template drift) both fail identically with my 13 source files stashed — pre-existing, not branch-caused.

`npx tsc --noEmit -p packages/squad-cli/tsconfig.json` clean. `eslint` on all 15 changed files: 0 errors, 107 pre-existing `no-console`/`n/no-sync` warnings matching each file's existing style, nothing new.

Diff hygiene: `git diff` vs `git diff -w` identical on every file (0 lines difference), including the two files that turned out to have mixed CRLF/LF/BOM conventions on this box (`cleanup.ts` is CRLF+BOM, most of the rest LF) — verified each against `git show HEAD:<file>` directly rather than the working-tree copy, since this session hit a real discrepancy where several files read back as CRLF on disk despite being LF in the actual git blob.

---

### ⚠️ Quick Check
- [x] Changeset added: `.changeset/fix-1490-watch-external-state.md` (patch `@bradygaster/squad-cli`)

#### Branch & Commit
- [x] Verified diff contains only intended changes (8 source + 6 test files + changeset)
- [x] PR is not in draft mode

#### Build & Test
- [x] `npm run build -w packages/squad-cli` passes
- [x] `npm test` — targeted suites 97/98 (1 pre-existing unrelated failure); full-suite baseline compared against unmodified `dev`
- [x] `npm run lint` passes (tsc clean)
- [x] `eslint` — 0 errors

#### Changeset
- [x] Changeset added (cli patch)

#### Docs
- [x] N/A — internal capability wiring, no public CLI surface changed

#### Exports
- [x] `discoverSubSquads`/`SubSquad` newly exported from `watch/index.ts` for testability, matching the existing `reportBoard` pattern in the same file — not a new module, no subpath export change

---

### Breaking Changes
None. `WatchContext.stateRoot` is a new required field, but the only two places that construct a `WatchContext` from scratch (`runWatch`, `runLoop`) both already had `stateDir` in scope and now populate it. External capabilities implementing `WatchCapability` receive a richer context, not a different one.

### Waivers
None.


omercangumus:
CI's test job failure is unrelated — `upgrade.test.ts` still hardcodes `casting_policy_version: '1.1'`, #1643 bumped the template to `1.2` earlier today and missed that assertion. Confirmed identical failure with this PR's diff fully stashed against unmodified dev. Filed #1661, one-line fix, not touching it here to keep this PR scoped to #1490.

Issue #1661, opened 2026-08-08
test: upgrade.test.ts still expects casting_policy_version 1.1, template bumped to 1.2 in #1643


omercangumus:
## Summary

\`test/cli/upgrade.test.ts\` > \`ensureCastingDefaults scaffolds registry.json, policy.json, history.json\` hardcodes \`casting_policy_version: '1.1'\`. #1643 bumped every \`casting-policy.json\` template mirror to \`1.2\` (custom universes / descriptive default naming) but didn't touch this assertion. Fails deterministically on \`dev\` right now, not flaky — hit it on CI for #1660, confirmed identical failure with that PR's diff fully stashed.

## Repro

\`\`\`
npx vitest run test/cli/upgrade.test.ts -t "ensureCastingDefaults scaffolds registry"
\`\`\`

\`\`\`
FAIL  test/cli/upgrade.test.ts > CLI: upgrade command > ensureCastingDefaults scaffolds registry.json, policy.json, history.json
AssertionError: expected { casting_policy_version: '1.2', …(4) } to have property "casting_policy_version" with value '1.1'
 ❯ test/cli/upgrade.test.ts:542:20
    540|     expect(registry).toHaveProperty('agents');
    541|     const policy = JSON.parse(readFileSync(join(dir, '.squad', 'castin…
    542|     expect(policy).toHaveProperty('casting_policy_version', '1.1');
\`\`\`

## Fix

One-line: \`test/cli/upgrade.test.ts:542\`, \`'1.1'\` → \`'1.2'\`.

bradygaster:
Already fixed on dev — assertion at line 542 already expects `1.2`. Test passes.

PR #1665, opened 2026-08-09
fix(cli): verify process identity before killing tracked watch PIDs

### LINKED ISSUE #1641: bug: watch's pid-tracker kills any process matching a tracked PID, no identity check
## Summary

`PidTracker.cleanupStale()` (`packages/squad-cli/src/cli/commands/watch/pid-tracker.ts:84-106`) kills every PID in `.watch-pids` that's still alive, using nothing but `process.kill(pid, 0)` to decide "still alive" and `killTree(pid)` to kill it. `TrackedProcess` stores `name` and `spawnedAt`, but neither is ever compared against the real process before the kill — they're only used in the log line after the kill already happened. Same for `cleanup()` (line 65-80), same mechanism.

**Proven** (verified below, this run): the kill path is bare PID-number equality with zero identity check — no cmdline comparison, no start-time comparison, no owner check. A tracked record can claim a spawn time a month in the past for a process that actually started 90 seconds ago, and `cleanupStale()` kills it anyway without any contradiction check.

**Reasoned, not simulated on this machine**: on a real crash where watch dies without cleaning up its `.watch-pids` file, followed by a reboot (or just enough process churn for the OS to cycle back through the PID space), the OS can reassign that same PID number to a completely unrelated process before watch restarts and calls `cleanupStale()`. Given the proven behavior above, that unrelated process would be killed. I can't trigger a real PID-space wraparound + reassignment on this machine to observe that exact end-to-end chain, so I'm not claiming to have reproduced *that* half — only that the code has no defense against it if it happens.

## Steps to reproduce

Verified on `dev` @ a4b60e8f, Windows 11, Node v22.21.0.

Started a real, long-running process to stand in for "some unrelated process that now happens to occupy a given PID":

```js
node -e "console.log('sleeper started, pid=' + process.pid, 'startedAt=' + new Date().toISOString()); setTimeout(()=>{}, 300000);"
// sleeper started, pid=26820 startedAt=2026-08-07T23:19:00.128Z
```

Confirmed independently via the OS, not just the process's own claim:

```
PS> Get-Process -Id 26820 | Select-Object Id,ProcessName,StartTime
   Id ProcessName StartTime
   -- ----------- ---------
26820 node        8.08.2026 02:19:00
```

Wrote a `.watch-pids` file claiming this PID belongs to an entry from a month ago:

```json
[
  {
    "pid": 26820,
    "name": "totally-unrelated-agent-spawn-from-last-week",
    "spawnedAt": "2026-07-01T00:00:00.000Z"
  }
]
```

Ran `PidTracker.cleanupStale()` directly against that team root (compiled module, no mocks):

```
ALIVE BEFORE: true
⚠️ Killed orphaned process: totally-unrelated-agent-spawn-from-last-week (PID 26820, spawned 2026-07-01T00:00:00.000Z)
cleanupStale() returned: 1
ALIVE AFTER: false
```

The process's real start time (`2026-08-08 02:19:00`, confirmed by `Get-Process`) flatly contradicts the tracked record's claimed `spawnedAt` (`2026-07-01`, over a month earlier). `cleanupStale()` never checks this — it killed the process purely because the PID number matched.

## Root cause

`packages/squad-cli/src/cli/commands/watch/pid-tracker.ts`:
- `isAlive()` (line 38-45): `process.kill(pid, 0)` — existence check only, no identity.
- `killTree()` (line 48-62): kills by PID, no identity check.
- `cleanupStale()` (line 84-106) and `cleanup()` (line 65-80): both call `isAlive` then `killTree` in sequence with no code path that reads `proc.name` or `proc.spawnedAt` before killing — those two fields exist only for the post-kill log message at line 96.

## Suggested fix direction

Before killing a PID from a stale/tracked record, cross-check something about the live process against what was recorded — process start time is available cross-platform (`wmic process where ProcessId=X get CreationDate` / `/proc/<pid>/stat` starttime / `ps -o lstart=`) and would catch exactly the contradiction demonstrated above. If the live process's actual start time doesn't match `spawnedAt` (within some tolerance), skip the kill and log a warning instead of silently trusting the PID number.

omercangumus:
### What
`squad watch`'s `PidTracker` now cross-checks a live process's real OS start time against the tracked `spawnedAt` before killing it in `cleanup()`/`cleanupStale()`, instead of trusting a bare PID number.

### Why
Fixes #1641

`isAlive()` is `process.kill(pid, 0)` and `killTree()` kills by PID alone — `name`/`spawnedAt` were only ever used in the post-kill log line, never checked before the kill. If watch dies without cleaning up `.watch-pids` and the OS reassigns that PID to something unrelated before the next `cleanupStale()` run, the tracker kills whatever now owns that number. Repro is in the issue: a `.watch-pids` entry claiming a month-old spawn gets killed against a process that actually started 90 seconds ago, no contradiction check anywhere in the path.

### How
- `getProcessStartTime(pid)`: looks up the OS's real start time for a PID — `wmic process where ProcessId=X get CreationDate` on Windows (parses the CIM_DATETIME UTC-offset field properly, not just the local wall-clock digits), `ps -o lstart= -p X` elsewhere. Returns `null` if it can't be determined.
- `matchesTrackedIdentity(proc)`: compares that against `proc.spawnedAt` with a 5-minute tolerance (spawn/scheduling delay, not the minutes-to-months drift a real PID reuse produces). Fails open (returns `true`) when the start time can't be determined, so normal cleanup isn't blocked when `wmic`/`ps` aren't available.
- `cleanup()` and `cleanupStale()`: both call the check right before `killTree()`; on mismatch they log a warning and skip the kill instead of proceeding.

### Testing
Added a `identity check (#1641)` block to the existing `test/cli/watch-pid-tracker.test.ts` suite, same mock style as the rest of the file (execFileSync mocked at module level, `process.kill` spied per-test):
- skips the kill when the OS start time contradicts the tracked record (recreates the issue's exact repro numbers)
- still kills when the start times line up within tolerance
- still kills when the start time can't be determined at all (fail-open, matches prior behavior for anyone without `wmic`/`ps`)
- parses a real Windows `wmic` `CreationDate` response and a real POSIX `ps -o lstart=` response into the correct UTC instant

```
npx vitest run test/cli/watch-pid-tracker.test.ts
 Test Files  1 passed (1)
      Tests  16 passed (16)
```

Ran the broader watch suite too — `test/cli/watch.test.ts` hit its known cold dynamic-import 5s timeout (3 tests), confirmed pre-existing by rerunning the same file with `--testTimeout=20000`, all 6 pass. Not caused by this change; same quirk documented on prior PRs against this repo.

`cd packages/squad-sdk && npm run build` then `cd packages/squad-cli && npm run build` (both package-local, no root prebuild churn) — clean, `tsc --noEmit -p packages/squad-cli/tsconfig.json` clean. `eslint` on both changed files: 0 errors, 12 warnings, all `n/no-sync`/`no-console` on the same categories the file already triggered before this change (`execFileSync`, `console.log`) — no new warning categories.

Diff hygiene: `git diff` vs `git diff -w` identical, `git diff --check` clean — hit the usual CRLF drift on this box (working copy silently went CRLF on both edited files despite LF-committed blobs), normalized back to LF and reverified against `git show HEAD:<path>` before staging.

---

### ⚠️ Quick Check
- [x] Changeset added: `.changeset/1641-pid-tracker-identity-check.md` (patch `@bradygaster/squad-cli`)

#### Branch & Commit
- [x] Branch created from `dev` (upstream, fresh fetch)
- [x] Verified diff contains only intended changes (1 source + 1 test file + changeset)
- [x] PR is not in draft mode

#### Build & Test
- [x] `npm run build` (per-package) passes
- [x] `npm test` — targeted suite 16/16; broader watch suite checked, only the known pre-existing cold-start timeout
- [x] `tsc --noEmit` clean
- [x] `eslint` — 0 errors, pre-existing warning categories only

#### Changeset
- [x] Changeset added (cli patch)

#### Docs
- [x] N/A — internal behavior fix, no public CLI surface changed

#### Exports
- [x] N/A — no new modules or exports

---

### Breaking Changes
None. `matchesTrackedIdentity` is additive and private; the only externally visible change is that a kill can now be skipped (with a logged warning) in the specific case where the live process contradicts the tracked record.

### Waivers
None.


omercangumus:
the one CI test failure is pre-existing, unrelated to this PR — `test/cli/upgrade.test.ts > ensureCastingDefaults scaffolds registry.json, policy.json, history.json` expects `casting_policy_version: '1.1'` but the template's on 1.2 now (bumped in #1643). that's #1661, already filed. this branch's diff is 3 files (pid-tracker.ts, its test, changeset) — nothing touches casting templates, and the version bump predates this branch on dev.

PR #1666, opened 2026-08-09
fix(sdk): wire FSStorageProvider rootDir in resolveSquadState, fix teamRoot='.' sentinel

### LINKED ISSUE #1555: FSStorageProvider constructed without rootDir in resolveSquadState() — state writes land in repo root instead of .squad/
## Summary

With `stateBackend: "local"` (which also covers the `"worktree"` backend), the `squad_state` MCP server writes all mutable state (decisions log, agent histories, memory index/audit, orchestration logs, etc.) to the **repository root** instead of `<repo>/.squad/`. This silently creates a full "shadow" state tree that diverges from the authoritative `.squad/` copy that agents and humans actually read with ordinary file tools.

## Root cause

`packages/squad-sdk/src/resolution.js` (compiled to `dist/resolution.js`), function `resolveSquadState()`:

```js
const stateStorage = backend.name === 'local'
    ? new FSStorageProvider()               // BUG: no rootDir
    : new StateBackendStorageAdapter(backend, paths.projectDir);
```

For the `local`/`worktree` backend, `FSStorageProvider` is constructed with **no root directory argument**. All relative keys used by the state tools (`"decisions.md"`, `"agents/<name>/history.md"`, `"log/<file>.md"`, `"memory/audit.jsonl"`, ...) then resolve against the MCP server process's current working directory (the repository root) rather than against `paths.projectDir` (the real `.squad/` directory).

Note that `paths.projectDir` is already computed and correctly passed on the other branch of the same ternary (`StateBackendStorageAdapter(backend, paths.projectDir)`), which strongly suggests this was an oversight rather than intentional behavior. Because zero constructor arguments are passed regardless of user configuration, **there is no configuration-level workaround** — this must be fixed in the SDK itself.

There is also a module-level instance a few lines above with the same unrooted construction (`const storage = new FSStorageProvider();`), which may be worth auditing for the same class of bug if it's used anywhere state keys are resolved relative to a project directory.

## Suggested fix

```js
const stateStorage = backend.name === 'local'
    ? new FSStorageProvider(paths.projectDir) // pass rootDir so relative keys resolve under .squad/
    : new StateBackendStorageAdapter(backend, paths.projectDir);
```

## Reproduction

1. Configure `stateBackend: "worktree"` (or `"local"`) with `teamRoot: "."` in a repo that has a `.squad/` directory.
2. Start the `squad_state` MCP server against that repo.
3. Call `squad_state_write` with key `log/probe.md` and any content.
4. Observe the file is created at `<repo-root>/log/probe.md`, **not** `<repo-root>/.squad/log/probe.md`.
5. As an immediate tell, `squad_state_list()` returns the repository's *top-level* listing (`.git`, `.github`, `.squad`, ...) instead of the contents of `.squad/`.

## Impact

Agents read governance state (e.g. `.squad/decisions.md`, `.squad/agents/*/history.md`) with ordinary file tools, while the MCP state server reads and writes a parallel shadow copy at the repo root. The two records diverge silently — neither side has any signal that the other exists. In our case this ran undetected for roughly 11 days and caused a real operational failure: an agent correctly refused to act on a governance decision because that decision, written via the state tool, was invisible in the file the agent actually reads.

Approximate drift found in one affected repo before reconciliation:
- ~95 shadow-only history entries for a single agent
- 22 shadow-only log files, plus several more with diverging content
- 23 shadow-only orchestration-log entries
- memory audit log: 4 → 15 entries after merging
- decisions archive: 210 → 732 lines after merging

The shadow paths were also **git-tracked and committed** by automated processes (a scheduled "watch" job), so the pollution entered version control history, not just the working tree.

## Detection

To check if another repo/instance is affected, look for a shadow tree at the repository root alongside `.squad/`: directories/files such as `agents/`, `log/`, `orchestration-log/`, `memory/`, `decisions.md` appearing at the repo root are the tell. `squad_state_health` will report `FSStorageProvider` as the active backend, and `squad_state_list()` returning `.git`/`.github`/other repo-root entries (instead of `.squad/` contents) confirms it live.

## Workaround until fixed

Patching the installed `node_modules/@bradygaster/squad-sdk/dist/resolution.js` copy in place works, but:
- It is **overwritten on the next `npm install` / `squad upgrade`** — it is not a durable fix.
- A **restart of the `squad_state` MCP server process is required** for the patch to take effect — the fix on disk does not retroactively affect an already-running Node process holding the old module in memory.

## Environment

- `@bradygaster/squad-cli` / `@bradygaster/squad-sdk`: `0.11.0`
- Node.js: `v22.15.0`
- OS: Windows
- Note: a stale npx-cache build resolving to `@bradygaster/squad-sdk@0.9.4` (pulled in via squad-cli's wildcard `"@bradygaster/squad-sdk": "*"` dependency range) was also found during investigation. That older build predates `resolveSquadState()` entirely and appears to be an unused, non-functional cache entry — flagged for awareness, not confirmed as a live risk, but the wildcard dependency range is presumably how such stale/mismatched builds can persist in npx caches at all.

omercangumus:
### What
`resolveSquadState()` now constructs the local-backend `FSStorageProvider` with `rootDir` set, and `resolveSquadPaths()` treats `config.teamRoot === '.'` as local mode instead of remote mode.

### Why
Fixes #1555

Two independent defects in the same path, both live on `dev`, both hit by `squad_decide`/`squad_state_*` (the MCP state tools wired up in `state-mcp.ts`):

1. `resolveSquadState()` built the local backend's storage as `new FSStorageProvider()` — no `rootDir`. `FSStorageProvider`'s traversal guard is `if (!this.rootDir) return filePath;` — an unset `rootDir` makes it a no-op, so it never validated anything. Every `squad_state_*`/`squad_decide` write just trusted whatever path got built upstream instead of confining it to `.squad/`.
2. `resolveSquadPaths()`'s remote-mode check is `if (config && config.teamRoot)`. `squad externalize` writes `teamRoot: '.'` into `config.json` as its "no separate team root" marker (`externalize.ts:99`) — but `'.'` is truthy, so this took the remote branch and computed `teamDir = path.resolve(projectRoot, '.')`, which is just `projectRoot` — one directory **above** `.squad/`, not `.squad/` itself.

Once (1) is fixed on its own, (2) turns into a live path-traversal rejection instead of a silent misplacement: `state-mcp.ts` builds `ToolRegistry(context.paths.teamDir, ...)`, every tool does `path.join(this.squadRoot, key)`, and with `teamDir` one level too high, those paths land outside the now-enforced `rootDir` — `squad_decide` starts throwing `Path traversal blocked: [team-root]\decisions\inbox\<file>.md`. Fixing both together is what actually gets state writes back to `.squad/decisions/inbox/` on an externalized project instead of either silently misplacing them or hard-failing.

### How
- `resolution.ts` `resolveSquadState()`: `new FSStorageProvider(paths.teamDir)` — matches what every local-backend caller already builds its paths against (e.g. `state-mcp.ts` passes `context.paths.teamDir` as `ToolRegistry`'s `squadRoot`).
- `resolution.ts` `resolveSquadPaths()`: `if (config && config.teamRoot && config.teamRoot !== '.')` — `'.'` now falls through to the existing local-mode branch (`teamDir: projectDir`), which is what it was always meant to mean.

### Testing
Added a `#1555 regression: rootDir wiring + teamRoot=. sentinel` block to `test/state-backend.test.ts`:
- `teamRoot: '.'` (the externalize.ts marker) resolves `mode: 'local'` and `teamDir === .squad/`, not the parent directory
- local-backend storage now rejects a path that escapes `teamDir` (`Path traversal blocked`) and still round-trips a path inside it — proves `rootDir` is actually wired, not just present
- an end-to-end run through `ToolRegistry` built the same way `state-mcp.ts` builds it (`new ToolRegistry(ctx.paths.teamDir, undefined, ctx.storage)`), calling the real `squad_decide` tool handler, asserting the decision file lands in `.squad/decisions/inbox/` and *not* one level up in the repo root

```
npx vitest run test/state-backend.test.ts
 Test Files  1 passed (1)
      Tests  147 passed (147)
```

Also ran the broader set of suites that touch `resolveSquadPaths`/`resolveSquadState` (dual-root resolver, effective-squad-dir, external-state, state-mcp, ensure-squad-path-dual, rc, copilot, export-import, extract, notes-promote, cast, consult, upgrade-state-backend, watch-notes-promote, mcp-spec-init, npm-registry-fallback):

```
Test Files  3 failed | 14 passed (17)
     Tests  3 failed | 343 passed (346)
```

All 3 failures (`copilot.test.ts`, `extract.test.ts`, `rc.test.ts`, each `module exports run* function`) are the same pre-existing cold dynamic-import 5s timeout as `watch.test.ts` hits on this box — confirmed by rerunning just those 3 files with `--testTimeout=20000`: all 58 tests pass. Not caused by this change.

`cd packages/squad-sdk && npm run build` then `cd packages/squad-cli && npm run build` (package-local, no root prebuild churn) — clean. `tsc --noEmit` clean on both packages. `eslint` on both changed files: 0 errors, resolution.ts's 43 warnings are the same pre-existing `n/no-sync` categories the file already had (0 new), state-backend.test.ts has 0.

Diff hygiene: `git diff` vs `git diff -w` identical, `git diff --check` clean — `resolution.ts` had drifted to CRLF on disk despite an LF-committed blob (same box quirk as usual), normalized back to LF and reverified against `git show HEAD:<path>` before staging.

---

### ⚠️ Quick Check
- [x] Changeset added: `.changeset/1555-fs-storage-rootdir.md` (patch `@bradygaster/squad-sdk`)

#### Branch & Commit
- [x] Branch created from `dev` (upstream, fresh fetch)
- [x] Verified diff contains only intended changes (1 source + 1 test file + changeset)
- [x] PR is not in draft mode

#### Build & Test
- [x] `npm run build` (per-package) passes
- [x] `npm test` — targeted suite 147/147; broader 16-suite regression sweep 343/346 (3 pre-existing unrelated failures, proven)
- [x] `tsc --noEmit` clean
- [x] `eslint` — 0 errors, pre-existing warning categories only

#### Changeset
- [x] Changeset added (sdk patch)

#### Docs
- [x] N/A — internal correctness fix, no public API surface changed (`resolveSquadState`/`resolveSquadPaths` signatures unchanged)

#### Exports
- [x] N/A — no new modules or exports

---

### Breaking Changes
None. Both changes make existing, documented semantics (rootDir confinement; `teamRoot: '.'` meaning "no separate team root") actually hold — no signature or behavior change for any caller that wasn't already relying on the broken behavior.

### Waivers
None.

### DIFF
--- a/.changeset/1555-fs-storage-rootdir.md
+++ b/.changeset/1555-fs-storage-rootdir.md
@@ -1,0 +1,5 @@
+---
+"@bradygaster/squad-sdk": patch
+---
+
+`resolveSquadState()` now constructs the local-backend `FSStorageProvider` with `rootDir` set to `paths.teamDir`, so its path-traversal guard actually validates state writes instead of no-op'ing on an unset rootDir. Also fixed `resolveSquadPaths()` treating a `config.teamRoot` of `"."` (the sentinel `squad externalize` writes) as remote mode, which pointed `teamDir` one level above `.squad/` — it now correctly falls through to local mode. Together these were breaking `squad_decide`/`squad_state_*` MCP tools on externalized projects with a "Path traversal blocked" error.
--- a/packages/squad-sdk/src/resolution.ts
+++ b/packages/squad-sdk/src/resolution.ts
@@ -365,8 +365,12 @@ export function resolveSquadPaths(startDir?: string): ResolvedSquadPaths | null 
   const isLegacy = name === '.ai-team';
   const config = loadDirConfig(projectDir);
 
-  if (config && config.teamRoot) {
+  if (config && config.teamRoot && config.teamRoot !== '.') {
     // Remote mode: teamDir resolved relative to the project root (parent of .squad/)
+    // '.' is the sentinel externalize.ts writes for "no separate team root" —
+    // falls through to local mode below instead of resolving to the parent
+    // of .squad/ (path.resolve(projectRoot, '.') === projectRoot, one level
+    // too high).
     const projectRoot = path.resolve(projectDir, '..');
     const teamDir = path.resolve(projectRoot, config.teamRoot);
     return {
@@ -847,8 +851,12 @@ export function resolveSquadState(startDir?: string, cliOverride?: StateBackendT
 
   // For local backend, use FSStorageProvider directly (more capable).
   // For git-notes/orphan, bridge via StateBackendStorageAdapter.
+  // rootDir is paths.teamDir (matches the squadRoot every local-backend
+  // caller — e.g. ToolRegistry in state-mcp.ts — builds its paths against),
+  // so the traversal guard actually validates instead of no-op'ing on an
+  // unset rootDir and letting a bad upstream path resolve silently.
   const stateStorage: StorageProvider = backend.name === 'local'
-    ? new FSStorageProvider()
+    ? new FSStorageProvider(paths.teamDir)
     : new StateBackendStorageAdapter(backend, paths.projectDir);
 
   return { paths, backend, repoRoot, storage: stateStorage };
--- a/test/state-backend.test.ts
+++ b/test/state-backend.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
-import { mkdirSync, rmSync, writeFileSync, readFileSync, existsSync } from 'node:fs';
+import { mkdirSync, rmSync, writeFileSync, readFileSync, existsSync, readdirSync } from 'node:fs';
 import { join } from 'node:path';
 import { execSync, execFileSync } from 'node:child_process';
 import { randomBytes } from 'node:crypto';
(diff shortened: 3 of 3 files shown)


omercangumus:
same CI failure as #1665 — pre-existing, unrelated: `test/cli/upgrade.test.ts > ensureCastingDefaults scaffolds registry.json, policy.json, history.json` expects `casting_policy_version: '1.1'`, template's on 1.2 (#1643). that's #1661. this branch is 3 files (resolution.ts, its test, changeset), nothing touches casting templates. 7373/7374 tests pass.

bradygaster:
Closing as a duplicate of #1695 — the diffs are byte-identical (same 3 files, same +61/-3 change to `resolution.ts`, `state-backend.test.ts`, and the `1555-fs-storage-rootdir` changeset). #1695 is rebased on current dev and is the one being merged. Thanks @omercangumus — the fix itself is landing, just via the other PR.

PR #1790, opened 2026-08-21
fix(ci): force LF on *.mjs so shebanged scripts stay loadable by vitest

### LINKED ISSUE #1788: check-changeset-drift.test.ts silently runs zero tests on Windows (shebang + CRLF)
## Summary

`test/scripts/check-changeset-drift.test.ts` **fails to load entirely on Windows checkouts** — it reports `SyntaxError: Invalid or unexpected token` and runs **0 of its 8 tests**. Green on Linux CI, so it has gone unnoticed since the file landed in #1481 (2026-07-19).

Discovered by Booster while working #1770 (PR #1786). Independently reproduced and root-caused here.

## Impact

The changeset-drift gate is a **CI enforcement gate** (`squad-ci.yml:202` runs `node scripts/check-changeset-drift.mjs`). Its unit tests are silently dead for every Windows developer. A local `npm test` shows a failure that reads like environment noise, so the natural response is to ignore it — meaning regressions in `evaluateDrift` / `listFragments` would not be caught locally.

This is the same failure *shape* the workstream keeps hitting: **something reports nothing and is mistaken for something passing.**

## Root cause

`scripts/check-changeset-drift.mjs` begins with a shebang:

```
#!/usr/bin/env node
```

It legitimately needs one — CI invokes it directly. But there is no `.gitattributes` rule for `*.mjs`, so it checks out **CRLF** on Windows. Vite/esbuild's shebang stripping does not survive the `\r`, leaving `#` as the first token of the module → `SyntaxError: Invalid or unexpected token`.

Node itself strips shebangs correctly, which is why `node -e "import('./scripts/check-changeset-drift.mjs')"` succeeds and masks the problem.

## Proof — three controlled runs

Working tree restored clean after each.

| Variant | Shebang | Line endings | Result |
|---|---|---|---|
| As committed | present | CRLF (113) | ❌ `SyntaxError`, **0 tests** |
| Shebang stripped | absent | CRLF | ✅ **8/8 pass** |
| **CRLF → LF, shebang kept** | present | LF | ✅ **8/8 pass** |

Natural control in the same directory — the three sibling suites in `test/scripts/` all import a `.mjs` too, and all pass:

| Test | Imports | First line | Result |
|---|---|---|---|
| `check-changeset-drift.test.ts` | `scripts/check-changeset-drift.mjs` | `#!/usr/bin/env node` | ❌ fails |
| `parse-diff.test.ts` | `scripts/impact-utils/parse-diff.mjs` | `/**` | ✅ passes |
| `risk-scorer.test.ts` | `scripts/impact-utils/risk-scorer.mjs` | `/**` | ✅ passes |

The shebang is the only differentiator. `test/scripts/` overall: `1 failed | 3 passed (4)`, 33 tests passing.

## Why CI never caught it

- The test file is in scope — `vitest.config.ts` has `include: ['test/**/*.test.ts']` and `npm test` is plain `vitest run`.
- Linux runners check out LF, so the shebang strips cleanly and all 8 tests pass.
- Latest `dev` CI run `32414111495` — `test` job **success**.
- File is **byte-identical to `origin/dev`**; this is pre-existing, not introduced by any in-flight PR.

## Suggested fix — one line

`.gitattributes` already encodes exactly this lesson for shell scripts:

```gitattributes
# Squad: shell scripts must stay LF — CRLF breaks the shebang on Linux/macOS.
*.sh text eol=lf
```

The same rule was simply never extended to `.mjs`:

```gitattributes
# Squad: CRLF breaks shebang stripping when a .mjs is imported by vitest on Windows.
*.mjs text eol=lf
```

Verified above: with LF and the shebang intact, all 8 tests pass.

Note this needs a re-normalize on existing checkouts (`git add --renormalize .`) for developers who already have the CRLF copy on disk.

## Success criteria

- [ ] `npx vitest run test/scripts/check-changeset-drift.test.ts` reports **8 passed** on a fresh Windows checkout
- [ ] `scripts/check-changeset-drift.mjs` retains its shebang and still runs via `node scripts/check-changeset-drift.mjs` in CI
- [ ] No other `*.mjs` regresses — `npx vitest run test/scripts/` reports 4 files / 41 tests passing
- [ ] A guard prevents recurrence: any `.mjs` imported by a test and carrying a shebang is pinned to LF

## Repro

```powershell
git checkout dev
npx vitest run test/scripts/check-changeset-drift.test.ts
# ❯ test/scripts/check-changeset-drift.test.ts (0 test)
# SyntaxError: Invalid or unexpected token
```

bradygaster:
## Root cause

`scripts/check-changeset-drift.mjs` starts with `#!/usr/bin/env node`. `.gitattributes` had no rule for `*.mjs`, so with `core.autocrlf=true` it checks out **CRLF** on Windows. Vite/esbuild's shebang stripping does not survive the `\r`, leaving a bare `#` as the module's first token -> `SyntaxError: Invalid or unexpected token`. `test/scripts/check-changeset-drift.test.ts` then loads **0 of its 8 tests**, silently. Linux CI checks out LF, so `dev` stayed green and this has been dead since #1481 (2026-07-19).

Node strips CRLF shebangs itself, which is why `node scripts/check-changeset-drift.mjs` and a plain `import()` of a `file://` URL both succeed and **mask** the defect. Only Vite's transform trips on it.

## Fix

`.gitattributes` already encodes this exact lesson for shell scripts (`*.sh text eol=lf` — "CRLF breaks the shebang"). It was never extended to `.mjs`:

```gitattributes
*.mjs text eol=lf
```

The shebang is **kept** — `squad-ci.yml:202` invokes `node scripts/check-changeset-drift.mjs` directly. Verified `node scripts/check-changeset-drift.mjs --mode=warn` still exits 0 post-renormalize.

`git add --renormalize -- "*.mjs"` lands in the same commit so existing checkouts converge rather than just the rule. **9 of 40** tracked `.mjs` blobs actually stored CRLF in the repo, so this was not a no-op.

Renormalize was scoped to `*.mjs` rather than `.` on purpose: `samples/**/*.sh` and `docs/pagefind.yml` churn CRLF<->LF in every worktree and a bare `--renormalize .` sweeps them in.

## Sweep — every tracked `.mjs` carrying a shebang (15)

```
packages/squad-cli/scripts/patch-esm-imports.mjs
packages/squad-cli/scripts/patch-ink-rendering.mjs
scripts/analyze-impact.mjs
scripts/build-standalone.mjs
scripts/bump-build.mjs
scripts/check-changeset-drift.mjs
scripts/check-exports-map.mjs
scripts/check-workflow-input-interpolation.mjs
scripts/measure-cold-start.mjs
scripts/measure-memory-value.mjs
scripts/promote-insider-tag.mjs
scripts/run-benchmarks.mjs
scripts/size-regression-guard.mjs
scripts/sync-skill-templates.mjs
scripts/sync-templates.mjs
```

All 15 are covered by `*.mjs text eol=lf`. Only `check-changeset-drift.test.ts` *statically imports* a `.mjs`; every other test reference shells out via `node <path>`, where Node's own shebang handling makes CRLF harmless. Shebangs also exist in 5 `.js`, 3 `.ts`, and 1 `.cjs` tracked file — none imported by a suite, left out of scope deliberately.

## Proof the test fails pre-fix

`test/scripts/mjs-shebang-loadable.test.ts` imports the real shebanged module through the bundler pipeline and replays every tracked shebang line (byte-exact, including its real terminator) as a fixture. Against a CRLF tree, **16/16 red with the production error**:

```
 ❯ test/scripts/mjs-shebang-loadable.test.ts (17 tests | 16 failed)
     × loads the real check-changeset-drift.mjs through the bundler pipeline
     × parses the shebang line of scripts/check-changeset-drift.mjs
     ... (13 more)

SyntaxError: Invalid or unexpected token
```

Getting this honest took two rejected mechanisms, both of which would have shipped a test that could not fail:

- `vite.transformWithEsbuild` throws in Vite 7 ("deprecated, migrate to transformWithOxc"), and `transformWithOxc` **does not reproduce** the bug — it normalizes CRLF->LF itself. Neither is a valid oracle.
- An earlier fixture specifier put the extension in the static part (`` `./${dir}/${name}.mjs` ``). That triggers `vite:dynamic-import-vars`, which pre-globs candidates at transform time and cannot see runtime-written fixtures — the 15 sweep tests failed with a *false* `Unknown variable dynamic import` instead of the real error. Switching to a fully opaque specifier routes through Vite's real module runner and produces the genuine failure above.

The guard asserts the **loaded/parsed outcome** — `evaluateDrift` and `listFragments` are callable functions — not the presence of a `.gitattributes` line.

## Verification

| Check | Result |
|---|---|
| `npx vitest run test/scripts/` before | `1 failed \| 3 passed (4)`, 33 tests |
| `npx vitest run test/scripts/` after | **5 passed (5), 58 tests** (41 required + 17 guard) |
| `node scripts/check-changeset-drift.mjs --mode=warn` | exit 0, shebang intact |
| `npm run build` | pass (6 prebuild-mutated files restored, not committed) |
| `npx eslint` on new test | clean |

Staged diffstat (11 files, deletion check empty):

```
 .gitattributes                             |    6 +
 docs/playwright.config.mjs                 |   26 +-
 docs/src/plugins/rehype-pagefind-attrs.mjs |   42 +-
 docs/tests/api-reference.spec.mjs          |  254 +++---
 docs/tests/astro-features.test.mjs         |    2 +-
 docs/tests/search.spec.mjs                 |  460 +++++-----
 scripts/capture-pr-screenshots.mjs         |   84 +-
 scripts/generate-api-docs.mjs              |  320 +++----
 scripts/pr-readiness.mjs                   | 1308 ++++++++++++++--------------
 scripts/sync-templates.mjs                 |  252 +++---
 test/scripts/mjs-shebang-loadable.test.ts  |   94 ++
```

`git diff --cached --ignore-cr-at-eol --stat` reduces to `.gitattributes` + the new test only — proving the 9 renormalized files carry **nothing but** line-ending changes. No `samples/**/*.sh` or `docs/pagefind.yml` staged.

Closes #1788

---

## Addendum: `*.snap` pinned to LF, and the lint's scope boundary written down

EECOM found that `npm test` dirties a second tracked file: `test/__snapshots__/parser-contracts.test.ts.snap`. Vitest writes snapshots with LF unconditionally; `*.snap` had no rule, so it checks out CRLF on Windows. Added `*.snap text eol=lf`.

**Reproduced red before fixing** (this is the same bar as the rest of the PR):

```
$ git status --porcelain              # clean
$ npx vitest run test/parser-contracts.test.ts
  Test Files  1 passed (1)
       Tests  16 passed (16)
$ git status --porcelain
 M test/__snapshots__/parser-contracts.test.ts.snap     <-- 16 green tests, dirty tree
```

With the rule, the same run leaves the tree clean.

### It does not ride the renormalize — correcting my own first assumption

I expected this to need `git add --renormalize` like the `.mjs` files did. It does not, and the measurement says why:

| | CRLF count | bytes |
|---|---|---|
| committed blob | **0** | 2991 |
| Windows worktree | **151** | 3142 |

The blob is **already LF**. `git add --renormalize -- "*.snap"` stages nothing — verified, empty diffstat. This is the mirror image of the `.mjs` case, where the blobs were CRLF and renormalize was the whole fix.

What an existing Windows checkout needs is for the working file to be **rewritten as LF**. A CRLF worktree under the new rule still reports ` M` (git diff is empty; it's git telling you it will replace the CRLF next time it touches the file). The first `npm test` after this merges does that rewrite by itself, and the tree stays clean from then on. **No manual remediation step.**

### Scope boundary — what this lint does not catch

`scripts/check-shebang-eol.mjs` would have been **green** the entire time the snapshot file was dirty, because a `.snap` file has no shebang. That limit is now written into the lint's own header rather than left to be rediscovered:

| Class | Symptom | Example | Lint catches? |
|---|---|---|---|
| **1 — strict parse** | file fails to load | `.mjs` CRLF shebang → `SyntaxError`, zero tests run | ✅ yes |
| **2 — tool rewrite** | tree perpetually dirty; a broad `git add` commits noise | `.snap` written LF, checked out CRLF | ❌ **no** |

Class 1 has an exact, cheap, static signature — the first two bytes are `#!`. Class 2 does not: "files some tool writes with LF" cannot be determined by reading the file. Extending the lint to guess would trade a sound check for an unsound one, so the boundary is deliberate and documented. New generated or tool-written file types get pinned in `.gitattributes` by hand.

Class 2 is the one with the data-loss fuse — it is the shape where a broad `git add` sweeps up an unrelated real change sitting in the same tree.


bradygaster:
Applied `skip-changelog`: this PR touches `packages/squad-cli/src/cli-entry.ts`, but only to rewrite its CRLF shebang to LF. `git diff --ignore-cr-at-eol` reduces the file to zero changed lines, and `tsc` already normalized the shebang on emit, so no published artifact or user-visible behavior changes. A changelog entry would announce a fix nobody experienced — the escape hatch is the honest option here rather than manufacturing release-note signal.

PR #1805, opened 2026-08-21
fix(ci): enumerate the EOL blob check by rule, not by shebang

### LINKED ISSUE #1804: check-shebang-eol gate is blind to the churn class it exists to catch (2 CRLF blobs live on dev)
## The gap

`scripts/check-shebang-eol.mjs` (added in #1790) enumerates candidate files by **`#!` at byte 0**, then checks two things about them: that a `.gitattributes` rule pins them to LF, and that the committed blob doesn't store CRLF.

But the bug class it exists to prevent is not defined by shebangs. It's defined by **`eol=lf` rule + CRLF blob**. Those two sets overlap; **neither contains the other.**

The consequence is a live false all-clear. Measured on `dev` @ `dbb4c33b`:

```
tracked files              : 1811
files with eol=lf attribute:  173
  of those, CRLF blob      :    2
     - docs/src/pages/rss.xml.js               (no shebang: gate BLIND)
     - packages/squad-cli/src/remote-ui/app.js (no shebang: gate BLIND)
```

Both are `.js`, covered by the `*.js text eol=lf` rule added in #1790 — and neither was renormalized when the rule landed. The gate reports:

```
Shebang EOL check passed: all 45 shebanged files are pinned to LF.
EXIT=0
```

## Why this half is the dangerous half

A rule over a CRLF blob means git normalizes the working tree to LF while the blob never moves. **The diff can never close.** The file shows modified in every worktree forever, `git restore` doesn't stick, and a broad `git add` sweeps it up — along with whatever real change happens to share the tree. Only `git add --renormalize` clears it.

Nothing about it is fatal, which is why it survives for years. This is the same condition that made `samples/**/*.sh` and `docs/pagefind.yml` perpetually dirty and was misread all session as inert churn.

Scale is irrelevant to the failure: `rss.xml.js` has **one** CRLF line. It has presumably sat there for as long as `*.js text eol=lf` has existed, producing a permanent one-line diff nobody could close by the obvious command.

## Fix

Give each invariant its own enumeration, as a union rather than a replacement:

| Invariant | Enumerate over | Why that set |
|---|---|---|
| `UNPINNED` | files starting with `#!` | only the `#!` walk finds a shebanged file with no rule |
| `CRLF-BLOB` | files where `git check-attr eol` = `lf` | the churn class is defined by the rule, not by the shebang |

Then renormalize the two files. Both changes are pure EOL — `git diff --ignore-cr-at-eol` is empty.

## Scope boundary (to remain documented, not fixed)

Still not covered: a file that *ought* to be pinned, has no shebang, and has no rule yet — vitest snapshots were exactly this until `*.snap text eol=lf` was added. There is no cheap static signature for "some tool writes this file with LF"; guessing would trade a sound check for an unsound one. New tool-written types get pinned by hand, and `CRLF-BLOB` enforces them from the moment a rule exists.

## Measurement note

Count CRs by reading blob bytes directly (`git cat-file` into a Buffer), or compare blob SHAs. Piping `git cat-file` through PowerShell re-encodes line endings and yields a confidently wrong number.

Follow-up to #1788 / #1790.

bradygaster:
Closes #1804. Follow-up to #1788 / #1790.

## Root cause

`scripts/check-shebang-eol.mjs` enumerated candidate files by **`#!` at byte 0**, then asked two questions about that one set: *is it pinned to LF*, and *does its blob store CRLF*.

The second question has the wrong enumeration. The churn class is defined by **`eol=lf` rule + CRLF blob** -- not by shebangs. The two sets overlap and **neither contains the other**, so the gate was blind to every pinned file without a `#!`, which is most of them.

## The blindness was live on `dev`

Measured at `dbb4c33b`:

```
tracked files              : 1811
files with eol=lf attribute:  173
  of those, CRLF blob      :    2
     - docs/src/pages/rss.xml.js               (no shebang: gate BLIND)
     - packages/squad-cli/src/remote-ui/app.js (no shebang: gate BLIND)
```

Both are `.js`, both covered by the `*.js text eol=lf` rule **added in #1790**, and neither was renormalized when that rule landed. So the rule created the condition while the gate it shipped alongside reported:

```
Shebang EOL check passed: all 45 shebanged files are pinned to LF.
EXIT=0
```

A/B against the same tree, old gate vs new:

| | result |
|---|---|
| gate as merged on `dev` | `passed: all 45 shebanged files are pinned to LF` - **EXIT=0** |
| widened gate | `FAILED: 2 problem(s)`, both `CRLF-BLOB` - **EXIT=1** |

## Why this half is the dangerous half

A rule over a CRLF blob means git normalizes the working tree to LF while the blob never moves, so **the diff can never close**. Modified in every worktree forever, `git restore` does not stick, and a broad staging command sweeps it up along with whatever real change happens to share the tree. Only `--renormalize` clears it.

Nothing about it is fatal, which is exactly why it survives for years. `rss.xml.js` has **one** CRLF line -- the failure mode in miniature, and proof it is scale-independent.

## Fix -- a union, not a replacement

Each invariant now walks its own set:

| Invariant | Enumerated over | Why that set |
|---|---|---|
| `UNPINNED` | files starting with `#!` | only the `#!` walk finds a shebanged file that no rule pins |
| `CRLF-BLOB` | files where `git check-attr eol` = `lf` | the churn class is defined by the rule |

Also: `eolAttributes` now feeds paths over **stdin** rather than argv. It is called with every tracked file now (~1800 paths, >70KB), which overflows the 32767-character Windows command-line limit.

Both offending files renormalized. **Pure EOL** -- `git diff --cached --ignore-cr-at-eol --stat` across the pair is empty, zero content lines.

## Proof the tests fail pre-fix

Swapped in dev''s pre-widening `check-shebang-eol.mjs`, kept the new tests:

```
 FAIL  > flags a NON-shebang file pinned to LF whose blob stores CRLF
   - [ ObjectContaining{"file": "web/app.js", "kind": "crlf-blob"} ]
   + []

 FAIL  > scans far more LF-pinned files than shebanged ones (guards the widened enumeration)
 TypeError: Cannot read properties of undefined (reading 'length')
 > test/scripts/check-shebang-eol.test.ts:191:26
     191|     expect(result.pinned.length).toBeGreaterThan(100);

 Tests  2 failed | 9 passed (11)
```

Restored the fix -- **11 passed**. The non-shebang test asserts its fixture is *absent* from the shebang enumeration before asserting the violation, so it cannot pass by accident if the scan is ever narrowed back down.

## Scope boundary -- written into the file, not left to be rediscovered

Still **not** covered: a file that *ought* to be pinned, has no shebang, and has no rule yet. Vitest snapshots were exactly this until `*.snap text eol=lf` was added. There is no cheap static signature for "some tool writes this file with LF" -- it cannot be determined by reading the file -- so guessing would trade a sound check for an unsound one. New tool-written types get pinned by hand; from the moment a rule exists, `CRLF-BLOB` enforces it.

## Verification

```
$ git diff --cached --stat
 docs/src/pages/rss.xml.js               |   2 +-
 packages/squad-cli/src/remote-ui/app.js | 668 ++++++++++++++++----------------
 scripts/check-shebang-eol.mjs           | 124 ++++--
 test/scripts/check-shebang-eol.test.ts  |  41 ++
 4 files changed, 470 insertions(+), 365 deletions(-)

$ git diff --cached --diff-filter=D --name-only
(empty)

$ git diff --cached --ignore-cr-at-eol --stat -- docs/src/pages/rss.xml.js packages/squad-cli/src/remote-ui/app.js
(empty)
```

- `npm run build` -- passes. `prebuild` version/template churn restored, not committed.
- `npx vitest run test/scripts/` -- **7 files, 75 tests**, all passing.
- `node scripts/check-shebang-eol.mjs` -- `EOL check passed: 45 shebanged file(s) all pinned to LF, 173 LF-pinned file(s) all storing LF blobs.` EXIT=0.
- `node scripts/security-review.mjs` and `architectural-review.mjs` -- both clean locally, pre-checked because `dev` does not yet carry #1786 and any emitted finding would crash the base-branch reporter.

## Changeset

`packages/squad-cli/src/remote-ui/app.js` is touched, so `changelog-gate` fires -- but the change is **zero content lines**, purely CR removal in the stored blob. There is no behavior to announce and a release note would be manufactured signal. Requesting `skip-changelog` on the same reasoning accepted for `cli-entry.ts` in #1790.

## Measurement note

CR counts here come from reading blob bytes directly into a Buffer via `git cat-file`. Piping `git cat-file` through PowerShell re-encodes line endings and returns a confidently wrong number.


bradygaster:
**`skip-changelog` rationale.**

This PR touches `packages/squad-cli/src/remote-ui/app.js`, so `changelog-gate` fires. The change to that file is **zero content lines** -- purely CR removal from the stored blob:

```
$ git diff --cached --ignore-cr-at-eol --stat -- docs/src/pages/rss.xml.js packages/squad-cli/src/remote-ui/app.js
(empty)
```

No shipped behavior changes, so a changeset would announce a fix that no consumer experienced. That is manufactured release-note signal, which is the same species of false coverage this workstream has spent the night removing.

Same reasoning accepted for `cli-entry.ts` in #1790.

Issue #1731, opened 2026-08-17
Phase 0b: Add PR provenance marker and tighten implement-worker merge relay


bradygaster:
Parent: #1729

## Problem

Two defects in `workflows/squad-implement-worker.md` that the review gate depends on:

1. **No machine-readable provenance.** A Squad-authored PR is identifiable only by the `squad/implement-*` head branch prefix. Branch names are mutable and a bot login is ambiguous once other automation opens PRs. The reviewer needs a durable marker.
2. **Merge relay is under-constrained.** The `pull_request: [closed]` relay fires when `merged == true` and the head branch prefix matches, but does not verify the PR merged into the default branch. A merge into a scratch or stacked branch advances the epic as though the work had landed. The relay also drops `aw_context`, losing correlation between the merged PR run and the re-dispatch.

## Scope

`workflows/squad-implement-worker.md` only.

## Acceptance criteria

- [ ] Worker PR bodies include a provenance marker, e.g. `<!-- squad:implement issue=N run=<run_id> -->`, emitted on every created PR.
- [ ] Merge relay `if` condition additionally requires the PR base to equal the repository default branch.
- [ ] Relay preserves `aw_context` through the re-dispatch.
- [ ] Tests assert the marker format, the base-branch condition, and `aw_context` propagation.

## Notes

The relay dispatches with `target-ref: ${{ github.event.repository.default_branch }}` because merged head branches may be deleted. That is correct and should not change. This issue constrains *when* the relay fires, not what ref it targets.

bradygaster:
## Procedures triage brief — Phase 0b

### Owner & rationale
**Owner:** EECOM (squad:eecom). This lives in the implement-worker relay/provenance path and needs core workflow/runtime ownership rather than prompt-router ownership.

### Files in scope
Verified existing paths:
- workflows\squad-implement-worker.md
- 	est\gh-aw-implement-workflow.test.ts

### Prompt-budget impact
No direct workflows\squad.md impact: **+0 KB** to the router prompt. Router headroom remains **37,602 B** unless a later phase spends it.

### Dependencies
- **#1772 is closed**, so the shared merge-relay defect this phase depended on has landed.
- **#1732 is closed**, so compile/budget gating exists.
- No Tier 1 false-green issue blocks this phase; it is still sequenced before #1733 because #1733 consumes the provenance marker.

### Verification bar
[...]

Issue #1730, opened 2026-08-17
Phase 0a: Add issue-scoped concurrency and actor authorization to the /squad router


bradygaster:
Parent: #1729

## Problem

`workflows/squad.md` has no dispatcher-level concurrency and no actor authorization on mutating modes. Two `/squad implement` invocations on the same epic can race and double-dispatch the same child issue, producing duplicate `squad/implement-{N}-*` branches. Any actor who can comment on an issue can trigger dispatch, artifact-mutating acceptance, or activation.

## Scope

`workflows/squad.md` only.

## Acceptance criteria

- [ ] Router declares issue-scoped `concurrency` so concurrent invocations against the same issue serialize rather than race. Use `cancel-in-progress: false` — cancelling a partially-dispatched run leaves inconsistent state.
- [ ] Mutating modes (`implement`, `plan accept`, `plan activate`, and any mode that writes artifacts or emits `squad_implement_worker`) are guarded by an actor authorization check. Read-only modes (`research`, `plan` preview, status) stay open.
- [ ] Unauthorized invocation posts a clear refusal comment rather than failing silently.
- [ ] Tests assert both the concurrency block and the authorization guard.

## Notes

The existing dispatch guard already requires a non-empty numeric `issue_number` and forbids generic `dispatch_workflow`. This issue adds the two missing controls around that guard, it does not replace it.

bradygaster:
## Procedures triage brief — Phase 0a

### Owner & rationale
**Owner:** Procedures (squad:procedures). The core change is router policy: issue-scoped serialization plus actor authorization/refusal behavior in workflows\squad.md. Booster/FIDO should review the gate mechanics, but Procedures owns the prompt/control-plane shape.

### Files in scope
Verified existing paths:
- workflows\squad.md
- 	est\gh-aw-quality.test.ts

### Prompt-budget impact
Yes — this phase adds text to workflows\squad.md. Estimate **+1–2 KB** for concurrency metadata, authorization mode table/guard, and refusal text. Projected router size after this phase: **~63.4–64.4 KB**, leaving **~35.6–36.6 KB** under the 100 KB ceiling.

### Dependencies
- **#1732 is closed** and provides the compile/prompt-budget gate this phase needs.
- No Tier 1 false-green issue is a prerequisite for this phase; do **not** pull it forward unless Flight proves #1824/#1812/#1801/#1822/#1827/#1825 need it.

### Verification bar
[...]

Issue #1593, opened 2026-08-04
RFC: non-npm distribution for squad-cli — gh-aw activation shouldn't require registry.npmjs.org


tamirdresher:
@bradygaster — this comes out of #1587. You listed it as a known limitation and scoped it out, which was right for that PR, but I don't want to leave it sitting there unfiled.

> **npm required in the activation job** — works on GitHub-hosted runners (open egress); a self-hosted runner behind a corporate npm-blocking firewall would fail. Out of scope.

That's a hard blocker for anyone running gh-aw on self-hosted runners inside a corp network, which is a good chunk of the enterprise audience we keep saying we want. Today the only way to get Squad onto a machine is `npm i -g` or `npx`. If npmjs.org is blocked there is no plan B.

## Copilot CLI already solved this

Worth putting on the table before we design anything — the CLI we depend on ships five ways:

```
winget install GitHub.Copilot
brew install --cask copilot-cli
curl -fsSL https://gh.io/copilot-install | bash
npm install -g @github/copilot
```

...plus per-platform binaries on Releases. Their v1.0.78 has darwin/linux/linuxmusl/win32 × arm64/x64 tarballs, two MSIs, and a `SHA256SUMS.txt`.

The detail I found most interesting: `github-copilot-1.0.78.tgz` is **5.4 KB**, while `github-copilot-1.0.78-linux-x64.tgz` is **~71 MB**. It's the optionalDependencies platform-package trick that esbuild and swc use. npm is one front door to the artifacts, not the mechanism that produces them.

Which leaves us somewhere awkward: someone installs Copilot CLI via winget on a box with no Node anywhere, gets to Squad, and has to go install npm. **We are now the only npm dependency left in the stack.**

## Why this is more tractable than I assumed

I went through our dependency graph expecting to find a blocker and didn't:

- `node-pty` is the only native module, and it's already an **optionalDependency**
- `sql.js` is WASM, not native
- `node:sqlite` is a builtin (hence `engines: >=22.5.0`)
- `ink` / `react` are pure JS

So there's no native compile step anywhere in the required path. A Node SEA, `bun build --compile`, or the platform-package approach is a build-pipeline problem, not a rewrite.

[...]

PR #1818, opened 2026-08-21
fix(gh-aw): bump activation CLI pin 0.11.0 → 0.12.0 (new-repo readiness)

bradygaster:
## Summary

The Squad bootstrap in `workflows/shared/squad.md` pins the CLI version the activation job installs. The default fell two minor versions behind the npm registry, and the drift lands precisely on the population targeted by the upcoming new-repo E2E wave.

## Verified state before the bump

| Source | Version |
|---|---|
| activation default pin (`workflows/shared/squad.md` × 2, `docs/…/gh-aw.md` × 1) | `0.11.0` |
| `npm view '@bradygaster/squad-cli' dist-tags`.latest | `0.12.0` |
| `npm view '@bradygaster/squad-cli' dist-tags`.insider | `0.12.0` |
| in-repo `packages/squad-cli/package.json` | `0.13.0` (**unpublished** — `npm view @0.13.0` returns E404) |

An existing repo with `vars.SQUAD_CLI_VERSION` set is unaffected. A brand-new repo has no variable, falls through the `|| '0.11.0'` fallback in both the `env:` block and the `npx --yes …@${SQUAD_CLI_VERSION:-0.11.0}` one-liner, and lands on the two-minor-old CLI.

## Why `0.12.0` and not `0.13.0`

`0.13.0` is the version bump waiting on unmerged changesets (`.changeset/scheduler-quoted-task-ref.md`, `scribe-archival-integrity.md`, `max-reasoning-effort.md`, plus `README.md`). It is not on npm. Pinning there would take activation from stale to hard-broken — `npx --yes "@bradygaster/squad-cli@0.13.0"` returns E404 today, verified against the live registry.

## Why still a fixed pin, not a floating `latest` tag

Reproducibility over freshness, deliberately. The activation bootstrap runs in the unrestricted-network job and hands state off to the sandboxed agent job; a bad point release landing in the middle of that hop is a failure class that's hard to debug from an agent-side artifact. A fixed pin preserves the property that the same `@dev` commit produces the same activation. Consumers keep `vars.SQUAD_CLI_VERSION` as an opt-in override for freshness.

If we later want floating, it should be a separate deliberate decision, not fallout from this fix.

## Files changed

- `workflows/shared/squad.md` — 3 occurrences (comment on line 48, `env:` default on line 81, `npx` fallback on line 90). All bumped.
- `docs/src/content/docs/guide/gh-aw.md` — 1 occurrence (variable table on line 145).
- `git diff --cached --stat`: 2 files changed, 4 insertions(+), 4 deletions(-). No deletions.

## Changeset

**Not required.** The changelog gate regex at `.github/workflows/squad-ci.yml:271` matches:

```
^packages/squad-(sdk|cli)/src/|^packages/squad-(sdk|cli)/templates/|^\.squad-templates/|^templates/|^\.squad/agents/.+/charter\.md$
```

Neither `workflows/shared/` nor `docs/` matches. Rely on the live check to confirm; if the gate flags this I'll add one, but I don't expect it to.

## Out of scope, noted for later

- `docs/src/content/docs/features/standalone-install.md` still cites `v0.11.0` in curl-install examples, Actions defaults, and an MCP config sample (5 hits). That is the standalone/tarball distribution channel — a different concern from the gh-aw activation pin. Called out in the assessment; not touching it here to keep this PR surgical.
- Everything else from the new-repo readiness gap list (PR #1709 merge, `gh aw compile`+actionlint CI gate, `sync-squad-labels.yml` v7/v4 consistency, #1593, #1502, #1730, #1731) is parked pending Coordinator sequencing.

## Follow-up

Task 2 in the same coordinator request is empirical cold-start verification on a throwaway repo. That's a separate deliverable — findings will be reported after this PR is open. Naming will make disposability obvious.

## Verification checklist

- [x] npm dist-tags queried against live registry (`latest: 0.12.0`)
- [x] `npm view '@bradygaster/squad-cli@0.13.0'` verified as E404
- [x] Grepped all `.md`/`.mdx`/`.yml`/`.yaml`/`.ts`/`.js`/`.mjs`/`.cjs`/`.tsx`/`.jsx` for `SQUAD_CLI_VERSION` — only the 4 in-scope hits found
- [x] `git diff --cached --diff-filter=D --name-only` empty
- [x] `git status --short` shows only the two intended files
- [x] Known CRLF-noise files not touched (`docs/pagefind.yml`, `samples/storage-provider-azure/scripts/*.sh`, `packages/squad-cli/src/remote-ui/app.js`)

**Do not self-merge.** Requesting a reviewer per Coordinator direction.

### DIFF
--- a/docs/src/content/docs/guide/gh-aw.md
+++ b/docs/src/content/docs/guide/gh-aw.md
@@ -142,7 +142,7 @@ Set a repository variable to control which Squad CLI version the workflow uses:
 
 | Variable | Purpose | Default |
 |----------|---------|---------|
-| `SQUAD_CLI_VERSION` | Squad CLI version to install during activation | `0.11.0` |
+| `SQUAD_CLI_VERSION` | Squad CLI version to install during activation | `0.12.0` |
 
 Set it in **Settings → Secrets and variables → Actions → Variables**.
 
--- a/workflows/shared/squad.md
+++ b/workflows/shared/squad.md
@@ -45,7 +45,7 @@
 #
 # Optional custom Squad CLI version:
 #   vars.SQUAD_CLI_VERSION
-#   Default is 0.11.0.
+#   Default is 0.12.0.
 #
 # Optional model override:
 #   vars.SQUAD_MODEL
@@ -78,7 +78,7 @@ jobs:
 
       - name: Initialize Squad team
         env:
-          SQUAD_CLI_VERSION: ${{ vars.SQUAD_CLI_VERSION || '0.11.0' }}
+          SQUAD_CLI_VERSION: ${{ vars.SQUAD_CLI_VERSION || '0.12.0' }}
           GH_TOKEN: ${{ steps.squad-app-token.outputs.token || secrets.SQUAD_GITHUB_TOKEN || github.token }}
         run: |
           # Preserve committed cast state: if .squad/team.md already exists with
@@ -87,7 +87,7 @@ jobs:
             echo "✓ Existing squad team detected with roster entries — skipping init."
           else
             echo "No existing squad team found — running squad init."
-            npx --yes "@bradygaster/squad-cli@${SQUAD_CLI_VERSION:-0.11.0}" init --preset default --state-backend local
+            npx --yes "@bradygaster/squad-cli@${SQUAD_CLI_VERSION:-0.12.0}" init --preset default --state-backend local
           fi
 
       - name: Upload Squad state artifact


bradygaster:
## ✅ Approved — Flight (Lead / Architect)

Correct, minimal, and empirically validated. No defects. Approving and squash-merging.

### What I verified independently

| Claim | Result |
|---|---|
| 4 `SQUAD_CLI_VERSION` hits, all in-scope | ✅ Confirmed via repo-wide grep on `dev` |
| No changeset required | ✅ Regex at `squad-ci.yml:271` matches neither `workflows/shared/` nor `docs/`; live `Changeset Drift` check also passed |
| Scope: 2 files, no deletions, no CRLF sweep | ✅ Confirmed |
| Diff contents | ✅ Exactly 4 substitutions `0.11.0` → `0.12.0` — comment (L48), `env:` default (L81), `npx` fallback (L90), docs table (L145). Nothing else touched. |
| Live state | ✅ Re-checked, not trusted from handoff: `MERGEABLE` / `CLEAN`, single commit `e267cf01`, 16 checks pass |

**On `0.12.0` vs `0.13.0`:** npm's registry is unreachable from my environment, so rather than take the dist-tags on trust I corroborated from release evidence — `squad-npm-publish.yml` fires on `release: published`, so tags and npm are coupled. Highest release tag is **`v0.12.0`**; there is no `v0.13.0` tag; and in-repo `packages/squad-cli/package.json` is `0.13.0` with three pending unmerged changesets, i.e. the *next* version rather than a published one. **You picked correctly** — pinning to `0.13.0` would have taken activation from stale to hard-broken.

### One correction

> The default fell two minor versions behind the npm registry.

Against the registry it was **one** minor behind (`0.11.0` vs latest `0.12.0`). Two-minor only holds if you count the unpublished in-repo `0.13.0`. Narrative imprecision only — the code is right.

### On the fixed pin (answered for the coordinator)

Your reproducibility argument holds and I'd have made it: activation runs in the unrestricted-network job and hands state to the sandboxed agent job, so a bad point release landing mid-hop surfaces as a confusing agent-side artifact. Fixed is right.

But the pin needs a mechanism, and the history shows why. `git log -S` says the pin line has been touched **exactly once, ever** — at introduction in `60032b5b` on 2026-08-07, when `0.11.0` genuinely *was* latest (`v0.11.0` 2026-06-29, `v0.12.0` 2026-08-13). It was correct when written and became wrong with nobody touching it. That's structural decay, not sloppiness: every release silently re-breaks new-repo activation, and this one went 8 days undetected until your cold start caught it.

Recommending to the coordinator: a release-time bump PR, backstopped by a CI guard comparing the pin against the highest published release tag. Worth noting one trap for whoever picks that up — **do not** derive the pin from in-repo `package.json`; it reads the unpublished `0.13.0` and would automate us straight into the failure this PR avoids.

### On the `standalone-install.md` exclusion

Correct call, and I'd go a step further — the finding is slightly over-flagged. All five hits are illustrative rather than authoritative defaults, and the documented default is "latest release". Line 174 specifically is not a copy-pasteable config: it's prefaced by *"Normally that entry launches through `npx`: {…} **That would defeat the purpose here**"* and followed immediately by the bundle form the reader should actually use. It's the output the passage steers you away from. It doesn't undercut this fix. Separate low-priority docs pass; different distribution channel.

Nice work on the cold-start validation — reproducing the drift verbatim on a throwaway repo is what moved this from reasoned to proven.

Issue #1817, opened 2026-08-21
`.squad/` in `.git/info/exclude` silently hides new squad state files from git


bradygaster:
## Problem

`.git/info/exclude` (line 9) contains:

```
.squad/
```

commented as *"Squad consult mode (local only)"*.

This rule behaves asymmetrically, and the asymmetry is the hazard:

| File state | Behavior |
|---|---|
| Already tracked in git | Shows as modified, commits normally ✅ |
| **New / untracked** | **Invisible to `git status -uall`, silently never commits** ❌ |

An agent can write a file, see no error, report success, and have that file quietly die with the worktree. There is no signal at any point in the workflow.

`.git/info/exclude` is per-clone and not version controlled, so this is also invisible in code review and won't reproduce for other contributors — making it harder to diagnose when it does bite.

## Observed impact

During the agentic-workflows audit session (2026-08-21), Scribe wrote 8 files. 3 were already tracked and committed cleanly via #1816. The other **5 were silently dropped**:

- `.squad/log/2026-08-21T09-30-07-07-00-agentic-workflows-audit.md`
- `.squad/orchestration-log/2026-08-21T09-30-07-07-00-booster.md`
- `.squad/orchestration-log/2026-08-21T09-30-07-07-00-flight.md`
- `.squad/orchestration-log/2026-08-21T09-30-07-07-00-sims.md`
- `.squad/agents/sims/history.md` *(newly created — Sims had no prior history file, so this agent now has **no** persisted memory at all)*

The `sims/history.md` loss is the most consequential: it isn't a missing increment, it's a missing origin.

## Why this looks like a behavior change, not long-standing policy

`dev` currently tracks files that this rule would now prevent from ever being added:

- 4 files under `.squad/log/`
- 11 files under `.squad/orchestration-log/`

So these paths were previously committed as a matter of course. The exclude rule appears to have been added later, and it silently froze those directories — existing entries keep updating, new ones can never appear.

## Relevant context

- `.squad/` **is** tracked on `dev` by design and is stripped at release time by `squad-promote.yml` (`git rm -rf --cached --ignore-unmatch .squad/`), and `squad-preview.yml` fails if `.squad/` is tracked on a preview branch. So committing `.squad/` to `dev` does not leak into releases — the release path already handles it.
- The `Squad File Leakage` CI check (`scripts/check-squad-leakage.mjs`) is **informational only** — its header states *"Exit code: always 0 (informational only — does not block merge)"*. It would not have caught this.
[...]

Issue #1812, opened 2026-08-21
plan activate reads a hardcoded roster, not .squad/team.md — and reports a false provenance (re-diagnoses #1784)


bradygaster:
Found during the E4 staged walk, gate 8 (`plan activate`), run [`32471509974`](https://github.com/bradygaster/aspiregregator-squad-e2e/actions/runs/32471509974) on fixture `bradygaster/aspiregregator-squad-e2e`.

**This changes the root cause of #1784. #1784 is not retired — it is re-diagnosed.**

## Measured

The activation summary on fixture issue #22 states:

> **Roster set read from `.squad/team.md` (`## Members` → `Name` column):** `lead`, `reviewer`, `devrel`, `security`, `docs`

The fixture's actual `.squad/team.md` roster:

| Name | Role |
|---|---|
| Keaton | Lead / Architect |
| McManus | Backend Engineer (Orleans) |
| Fenster | Frontend Engineer (Blazor) |
| Hockney | Test Engineer |
| Kint | DevOps / Platform Engineer |

**None of `lead`, `reviewer`, `devrel`, `security`, `docs` appears anywhere in `team.md`.** The claim of provenance is false — the roster was not read from that file.

Reproduce:
```sh
gh api repos/bradygaster/aspiregregator-squad-e2e/contents/.squad/team.md \
  -H "Accept: application/vnd.github.raw"
```

## Corroboration

The fixture's existing `squad:*` labels:

```
squad
squad:lead
squad:devrel
squad:reviewer
```

`lead`, `devrel`, `reviewer` are a **strict subset of the falsely-reported roster**, and are exactly the three labels E3's `plan activate` minted (`squad:lead` ×3, `squad:devrel` ×1). They are not a subset of the real roster.

## Consequence — why #1784's diagnosis was wrong

The original framing was *"activate derives owner tokens from roles instead of using roster names,"* addressed as prompt token-salience in #1789.

The measurement says otherwise:

- **E3** — activate applied `squad:lead`/`squad:devrel` because those **are** members of the hardcoded list it actually uses. It was not leaking role tokens; it was correctly labelling against the **wrong roster**.
- **E4** — the implementation planner read `team.md` correctly and produced `Keaton`/`McManus`/`Fenster`/`Hockney`. Activate compared them against the hardcoded list, found no match, and **refused to apply any `squad:{agent}` label.**

[...]

Issue #1784, opened 2026-08-21
Owner/Agent binding fix (#1759) is ineffective live — model copies the forbidden tokens out of the prohibition


bradygaster:
## Summary

#1759's Owner/Agent binding fix (shipped in #1778) **does not work in the live system**. Experiment E3 confirmed it live: `/squad plan activate` minted four brand-new issues with forbidden role-string labels.

Worse — **the leaked values came from the prohibition itself**, not from the roster.

## Live evidence

Fixture `bradygaster/aspiregregator-squad-e2e`, issue #16, full long planning path (`research → triage → plan program → plan implementation → plan validate → plan accept scope → plan accept implementation → plan activate`). All 8 runs succeeded. Activate produced:

| Issue | Labels applied |
|---|---|
| #17 `[Epic] OPML Upload & Import` | `squad`, **`squad:lead`** |
| #18 `Parse OPML into subscription records` | `squad`, **`squad:lead`** |
| #19 `Import parsed feeds as subscriptions` | `squad`, **`squad:lead`** |
| #20 `OPML upload UI with results panel` | `squad`, **`squad:devrel`** |

Expected: `squad:keaton`, `squad:mcmanus`, `squad:fenster`, `squad:hockney`, `squad:kint`.

## This is not fixture staleness — ruled out explicitly

The fixture's `.github/workflows/squad.md` is **55,954 chars, byte-identical in length to `dev`**, and contains every #1778 marker:

- `Owner/Agent binding rule` ✅
- `never mint a role-derived label` ✅
- `Agent binding rule` ✅
- `Find Plan and Route` ✅
- `Must be ≥200 chars` correctly **absent** ✅

The fixture also has a valid `.squad/team.md` with a populated `## Members` table (Keaton, McManus, Fenster, Hockney, Kint).

**The model had the rule, had the roster, and disobeyed both.**

## Root cause: the negative examples are the source of the leaked values

`workflows/squad.md:913`:

> **Agent binding rule:** every `Agent` value MUST be a cast **Name** from the `Name` column of `.squad/team.md` (resolved via `.squad/routing.md`), never a Role string (`Lead`, `DevRel`) or lowercased role (`lead`, `reviewer`).

The model emitted `lead`, `lead`, `devrel`.

[...]

Issue #1779, opened 2026-08-20
Implement-worker slot refill searches only the immediate parent epic, stranding tasks in sibling epics


bradygaster:
## Origin

Surfaced by Procedures during the implementation of #1758 defect 2 (PR #1778). The dispatch side of defect 2 is fixed in that PR: `squad.md` now descends the sub-issue tree recursively and dispatches only leaf tasks, never Epics. The refill side — what the worker does *after* a task completes — is a separate responsibility that lives in `workflows/squad-implement-worker.md` (owned by EECOM, relevant to #1772). Procedures correctly did not touch that file.

## The Defect

`squad-implement-worker.md`'s "Continue Parent Epic After Merge" section refills the freed dispatch slot by looking at the **immediate parent epic's** remaining sub-issues. In a 2-level hierarchy (Root → Tasks) this is fine. In a 3-level hierarchy (Root → [Epic A, Epic B] → leaf Tasks), it fails:

1. Task under Epic A completes. Worker looks at Epic A's sub-issues for the next task.
2. Epic A's remaining tasks fill the freed slot — correct.
3. Epic A's last task completes. Worker looks at Epic A's sub-issues — finds none unstarted.
4. Worker exits without dispatching. **Epic B still has pending tasks, but the worker never looked there.**
5. The concurrency slots sit idle. No error is surfaced. The run completes green.

This is the same silent-no-op failure signature as #1772 — looks green, work is stalled.

## What Was Fixed in #1778 (not this issue)

`squad.md`'s activation dispatch now recursively walks the full sub-issue tree and hands out leaf tasks up to the 3-slot concurrency cap. That covers the *initial* distribution across all epics. This issue is exclusively about the *refill* behavior after a task merges.

## Failure Precondition

Requires a 3-level hierarchy with at least two epics, where one epic's tasks are fully dispatched before the other's. In the current `aspiregregator-squad-e2e` fixture (epic #5, single remaining child #9), this failure shape is not reachable tomorrow — S6 would need a fresh multi-epic root to trigger it. This is a known limitation Sims should record if S6 is run with a full 3-level hierarchy.

## Goal

After a task merges, the implement-worker dispatches the next unstarted leaf task from **anywhere in the root epic's sub-tree** — not only from the completing task's immediate parent epic — so that concurrency slots are never idle while work remains.

## Success Criteria

**SC-1 — Refill queries the full subtree, not the immediate parent.**  
[...]

PR #1819, opened 2026-08-21
docs(e2e): amend E4 for #1811/#1812, formalize E1 Rule D gate, add folder README

### LINKED ISSUE #1811: E4: document that Condition 0 must be evaluated first — it is the premature-read guard, not just a liveness floor
## Problem

`Condition 0` in `.squad/e2e/E4-agent-binding-verification.md` currently does **three** jobs, but only documents one:

| job | status |
|---|---|
| liveness floor — ≥1 issue created by `activate` | **designed and documented** |
| seed exclusion — census by *author*, excluding seed issue | patched in after the 48-second hole |
| **premature-read guard** | **undocumented and unowned** |

## Why the third job exists

The #1784 measurement is a **`safe_outputs` effect — job 5 of 6** in the gh-aw pipeline (`pre_activation`, `activation`, `agent`, `detection`, `safe_outputs`, `conclusion`).

`agent ✅` lands *before* the label state is written. Reading at that moment fails **in both directions**, depending on instrument:

| read prematurely | symptom |
|---|---|
| label census | **false RED** — the three pre-existing `squad:lead`/`squad:devrel`/`squad:reviewer` definitions look like a confirmed #1784 reproduction |
| timeline `labeled` events | **vacuous GREEN** — Condition 2 "passes" on an empty event list |

Switching instruments does not help: **the hazard is temporal, not instrumental.**

Condition 0 catches both, because a premature read yields zero Squad-authored issues → `INCONCLUSIVE` → stop before 1–3 are scored. **A premature read and a dead pipeline are indistinguishable at the census**, which is exactly why the same check covers both.

## Why this is fragile

This robustness is **accidental**. Nobody wrote it down. The condition *says* it is a liveness floor, so a future reader who can see the run is obviously alive will reasonably move it later or drop it as ceremony. **The timing guard then disappears silently and nothing fails** — both hazards only fire on an early read, which is rare and unlogged. It returns weeks later as a false red or a vacuous green with no connection to the edit.

## Fix

One sentence, living next to the condition it protects (not in a session log):

> Evaluating this first is what makes Conditions 1–3 readable. A premature read is indistinguishable from a dead pipeline here, so this check also guards against reading the label state before `safe_outputs` completes.

Plus a line in the capture procedure: **wait for all six jobs including `conclusion`; `agent ✅` is not sufficient.**

## Context

Found during the E4 staged walk (2026-08-21), gate 8 of 8. Raised by EECOM off a corrected relay about the pipeline's job count. The correction — *"I read a mid-run job list as the pipeline's shape"* — is what generated the hazard; the corrected fact alone would not have.

Owner: Sims (owns `.squad/e2e/E4-agent-binding-verification.md`). Not urgent for the 9am series — Condition 0 works today; this protects it from being refactored away.

bradygaster:
**Scope.** Documentation-only edits to `.squad/e2e/` authorized by the Coordinator after Sims' readiness assessment. No code changes, no fixture changes, no re-runs. Six items:

## 1. E4 — Phase 2 extended by one command

`plan activate` is now part of E4's scored path. It was the missing command in the n=1 run: `plan implementation` produced the correct `Agent` column (proving #1784's structural fix), but `plan activate` then minted labels off a **hardcoded roster** rather than the fixture's actual `.squad/team.md`. That's `#1812`. E4 was silent on it because activate wasn't executed.

## 2. E4 — Phase 3e roster-provenance gate

Two-part, order-insensitive whitespace-normalized set equality:
- **Gate A** — the activation summary comment MUST contain a provenance line (`Roster set read from … team.md`) naming its source. Absence is a fail.
- **Gate B** — the reported roster MUST equal the fixture's actual `## Members → Name` column, order-insensitive. If Gate A passes and Gate B fails, that reproduces `#1812` (right shape, wrong content). If Gate A fails at all, that's a separate silent-success bug.

Full PowerShell implementation lives in the doc. Runtime cost: +4 min, still unattended-safe.

## 3. E4 — contamination-table growth note

`⚠️ This table grows.` on Phase 3b. Each E4 run adds to the observed set; the table is n=1 today, but the next re-run will not start from empty state and the operator needs to know that up front.

## 4. E4 — Condition 0 dual-role callout (closes #1811 in-doc)

Condition 0 does three jobs — liveness floor, seed-exclusion census, and (the one that was undocumented) **premature-read guard**. Reading label state after `agent ✅` but before `safe_outputs` completes yields false RED for the label census and vacuous GREEN for timeline events. The doc now says: wait for all six pipeline jobs, not just `agent`.

## 5. E4 — stripped citations to nonexistent `squad-e2e-runbook.md`

Confirmed by the Coordinator with `git log --all --diff-filter=A -- '**/squad-e2e-runbook.md'`: that file has never existed on any ref. Four call-sites in E4 (Budget, Phase 0d, Phase 3a "gotcha #10", Follow-on #1779/#1772 discriminator, Related table row) were rewritten to be self-contained. Amendment log entry names the change.

## 6. E1 — Rule D promoted to a formal gate

`NOT VERIFIED (future E2 experiment)` is gone. Rule D is now a formal gate on E1 with three explicit verdicts (PASS/FAIL/NOT VERIFIED), a required-fields YAML escape hatch for `detection`-hang cases (blank fields are treated as FAIL — an escape hatch that accepts silence lets any run score NOT VERIFIED), and a PASS attribution rule: closure must be Squad-owned, not a human papering over the run. There will be no E2 — Rule D shares E1's trigger.

## 7. `.squad/e2e/README.md` — new file

Explains vestigial numbering: E2 was folded into E1 as Rule D, E3 was the pre-#1784 rehearsal that became E4's input rather than its own committed procedure, and `windows-test-baseline.md` is intentionally in this folder despite not being an E-scenario because operators running E-scenarios need it.

---

## Verification

Because `.git/info/exclude:9` contains `.squad/`, `git add` and `git commit` succeed silently for NEW `.squad/` files even when the blob never lands. All three files verified in the commit tree via `git cat-file -e $sha:$path` after commit:

```
OK  .squad/e2e/README.md                        4595 bytes
OK  .squad/e2e/E1-merge-continuation-relay.md   8622 bytes
OK  .squad/e2e/E4-agent-binding-verification.md 41502 bytes
```

Staged with explicit paths, no `git add .`/`-A`/`-a`. No known CRLF-noise files touched. Diff-stat: 3 files, 307+/61−. Zero deletions.

## What this PR is NOT

- No fixture refresh. Fixture was inspected as of `1b11c5f` for the assessment; refresh is Phase 0 of the next actual re-run.
- No `#1817` (`.git/info/exclude` trap) fix. A parallel session owns that investigation.
- No new E-scenario files. E2 is intentionally not being written.
- No re-run of E4. The n=1 verdict stands until the amended procedure is executed against a refreshed fixture. Phase 3e's regex anchors (`Roster set read from`, `team.md`, `Name`) are designed for detection today; they may need re-verification if the activation summary format changes.

## Related

- Closes `#1811` in-doc (verification requires an executed E4 re-run under the amended procedure).
- Provides scoring surface for `#1812` in-doc (same caveat).
- `#1817` handling documented in the readiness assessment; not addressed here.

Do not merge — reviewer routing pending.

### DIFF
--- a/.squad/e2e/E1-merge-continuation-relay.md
+++ b/.squad/e2e/E1-merge-continuation-relay.md
@@ -11,7 +11,7 @@
 
 The merge-continuation relay — which had **never once succeeded** before tonight — worked end to end on 2026-08-21.
 
-**E1 verdict: PASS** — all tested items passed. One item (epic closure / Rule D) was **not verified** due to an infrastructure hang, not a Squad defect.
+**E1 verdict: PASS (with one gate NOT VERIFIED)** — 7 of 8 tested items passed. The eighth — **Rule D / epic closure** — was **NOT VERIFIED** due to a `detection` infrastructure hang, not a Squad defect. Rule D is now a **formal gate on this procedure** (see *Rule D — the gate and its escape hatch* below), and any re-run must score it explicitly rather than treating an unobserved run as green.
 
 ---
 
@@ -88,4 +88,61 @@ Operational guidance: if the `detection` job has not completed within ~10 minute
 
 Epic closure was not re-tested after the fix. The session ended while the `detection` job hung. This is **not a Squad regression** — the infra hang prevented observation, not a logic failure.
 
-This item remains open for a future E2 experiment.
+Rule D is now a formal gate on E1 (see next section). It is **not** getting its own scenario number — an earlier draft of the E-series referred to a "future E2 experiment," but that scenario was never authored. Rule D shares E1's trigger (the merge-continuation relay closing the last leaf), so verifying it separately would either duplicate E1 or fabricate the state — neither is worth a distinct procedure. See `.squad/e2e/README.md` for the numbering history.
+
+---
+
+## Rule D — the gate and its escape hatch
+
+**Added post-execution, 2026-08-21.** Any re-run of E1 MUST score Rule D explicitly rather than defaulting to green on absence of contrary evidence.
+
+### The gate
+
+**Trigger.** All leaf tasks under an epic have been merged (their child issues CLOSED) and the merge-continuation relay has been observed to dispatch (i.e. E1's other 7 gates have passed on the final leaf).
+
+**Rule D verdict — three possible outcomes:**
+
+- ✅ **PASS** — the epic issue is CLOSED within a stated observation window (default: 10 minutes after the last leaf's merge-continuation relay run reaches `success`). The closure is attributed to Squad (either commented by the coordinator or effected via a Squad-authored PR/dispatch), not a manual close.
+- ❌ **FAIL** — all leaves are CLOSED, the observation window has expired, the epic is still OPEN, and no `Squad —` workflow run in that window shows a closure attempt. This is a real Rule D defect: closure logic exists but did not fire, or fired against the wrong target.
+- ⚠️ **NOT VERIFIED** — the observation was blocked by an infrastructure hang (`detection` on `Install ripgrep` is the known offender). Record the observation-blocking job's URL and step and the exact minute the run was cancelled. A NOT VERIFIED verdict is *not* a green: it is a documented absence of signal, and it does not close #1758.2 or any successor.
+
+### The escape hatch (required text on any NOT VERIFIED result)
+
+```yaml
+rule_d:
+  verdict: NOT VERIFIED
+  reason: "detection job hung >N min on step '<step name>'; observation window expired before closure could be witnessed"
+  blocked_by_job_url: "https://github.com/<owner>/<fixture>/actions/runs/<run_id>/job/<job_id>"
+  blocked_by_step: "<step name>"
+  cancelled_at_utc: "<yyyy-MM-ddTHH:mm:ssZ>"
+  observation_window_min: 10
+  window_started_utc: "<yyyy-MM-ddTHH:mm:ssZ>"      # when the relay run reached success
+  window_ended_utc:   "<yyyy-MM-ddTHH:mm:ssZ>"      # cancelled_at_utc for hang; +window_min otherwise
+```
+
+**Do not omit any of these fields on a NOT VERIFIED.** An escape hatch that accepts blank fields becomes a way to score anything as NOT VERIFIED. If any of the required text is missing, treat the result as **FAIL** (blocked-observation is a claim; unsubstantiated blocked-observation is silence).
+
+### PASS observation (the mandatory extractions)
+
+```powershell
+$FIXTURE = "bradygaster/aspiregregator-squad-e2e"
+$EPIC = <epic issue number>
+
+# The relay run's success timestamp — the start of the observation window.
+gh run list --repo $FIXTURE --workflow "Squad" --limit 20 `
+    --json databaseId,name,conclusion,updatedAt `
+    --jq '.[] | select(.conclusion=="success" and (.name|test("implement|relay"; "i"))) |
+         [(.databaseId|tostring),.name,.updatedAt] | join("  |  ")' 2>$null
+
+# Epic state at the end of the observation window.
+gh issue view $EPIC --repo $FIXTURE --json state,closedAt,timelineItems `
+    --jq '{state,closedAt,closer:(.timelineItems[]|select(.__typename=="ClosedEvent")|.actor.login)}' 2>$null
+```
+
+**PASS attribution rule.** `closer` must be a Squad-owned actor (`app/github-actions` posting a Squad workflow closure, or the coordinator's Squad-authored PR merging). A closure attributed to a human operator during the observation window is **not** a Rule D PASS — it is a Rule D FAIL that was manually papered over, and E1's whole point is that closure was supposed to be autonomous.
+
+### FAIL observation
+
+Record the full `gh run list --workflow "Squad"` output for the observation window along with the epic's `state` and `openIssues` count. A FAIL says the closure logic did not fire — that is the finding.
+
+---
--- a/.squad/e2e/E4-agent-binding-verification.md
+++ b/.squad/e2e/E4-agent-binding-verification.md
@@ -1,8 +1,21 @@
-# E4 — Agent-Binding Verification (#1784)
+# E4 — Agent-Binding Verification (#1784, #1812)
 
 > **Author:** Sims (E2E Test Engineer)
 > **Written:** 2026-08-21, from the E3 long-path rehearsal
-> **Status:** ✅ **EXECUTED 2026-08-21 — result recorded below.**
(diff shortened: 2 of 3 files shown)


bradygaster:
## ✅ Approved — Flight (Lead / Architect)

Merging. Four findings below are refinements, not blockers — none justify holding a +307/−61 net improvement to a document CI does not execute.

### Priority 1 — the 59 deletions: audited, zero procedural loss

Every deleted line traced to its fate. **No procedural step, gate, condition, or evidence-table row was lost.** Each deletion either survived in modified form, was relocated to a better place, or was correctly obsoleted by the scope extension.

Your framing of Sims' claim was slightly generous to them — it is *not* only "four citations plus a Related row." It is that, **plus** the epic-level owner-minting carve-out, **plus** several notes the scope change rendered obsolete. But every one of those extra removals is justified:

| Deleted | Fate |
|---|---|
| `## Scope limitation — what E4 does NOT cover` | **Renamed + widened** → `## Scope — what E4 covers, and what it does NOT` |
| Epic #17 / run `32433011339` / epic-level owner minting **not covered** | **Correctly obsoleted.** The carve-out existed because epic owners mint at `plan activate` — *downstream of where E4 stopped*. E4 now runs `plan activate`, and 3b checks every `squad:{x}` on a **newly created** issue, which includes epics. The uncovered surface became covered. |
| `~27 min` budget / E3 `26.7 min` derivation | **Retained and re-derived** → `~31 min`, with the `plan activate` delta measured live (`~2m43s`, run `32471509974`). The `~27 min` figure survives, correctly re-attributed to the *excluded* validate/accept path. |
| Gotcha ④ "runbook gotcha #1" | **Citation resolved** → self-contained silent-success explanation, expanded |
| "labels minted later… **weak corroboration only**" | **Correctly obsoleted** — labels *are* minted now that `activate` runs |
| "**The `Agent` column is authoritative**… that is a FAIL" | **Preserved, relocated** — subsumed into the three-way verdict table's FAIL row as a disjunction (`in Agent values, **or** on an issue newly created by this run`). Dirty column + clean labels still lands FAIL. Better placed than the original prose aside. |
| Evidence rows / verdict YAML | **All present**, expanded for #1812 (`fix_pr` → `fix_prs` map, `verdict_1812`, `gate_3e_*`) |
| `#1779`-vs-`#1772` discriminator "in the runbook" | **Citation resolved**, discriminator retained verbatim |
| `\| squad-e2e-runbook.md \| …` Related row | **Removed** — the row Sims claimed |

Runbook citations: old file had **exactly four** live citations (L292, L334, L642, L675). All four resolved. The only surviving mention in the new file is the amendment log *describing* the change — a meta-reference, not a live citation. **Claim verified.**

### Spec delivery: 7/7
All seven items delivered. Two exceeded scope in a good way: Condition 0 is documented as doing **three** jobs rather than the two asked for, and the 3b contamination note adds a re-derivation path (`gh label list` + `E4_START` cutoff) rather than just warning that the table grows. No E2 file created, as directed.

### A) Phase 3e fails **CLOSED** — but the doc says the opposite

The gate is real, not decoration:

```powershell
if (-not $claim.Success) { "GATE 3e (A): FAIL — no 'Roster set read from …' line" }
else { ...(B)... }
```

A reworded summary → no regex match → **A-FAIL is printed**, and `verdict_1812` becomes `INCONCLUSIVE`, never `PASS`. The `(?im)` flags without `(?s)` also correctly force all four anchor tokens onto a single line. **Fail-closed, as it should be.**

**Finding 1 — the caveat at L714-716 contradicts the code it describes.** It warns that *"a silently non-matching anchor scores green."* It does not; it scores A-FAIL. This understates the gate. Left as-is, it could lead a future operator to distrust a genuine A-FAIL, or worse, to "correct" the implementation *toward* fail-open. Fix the note to match the behaviour.

**Finding 2 — one genuine fail-open corner, and it is the exact class you're worried about.** In (B), `gh api … 2>$null` swallows failure. If the `team.md` fetch fails, `$b` is empty → `$realNames` is empty. If the provenance line is simultaneously present but captures a set that normalizes to zero entries, then `SetEquals(∅, ∅)` returns **true → B-PASS**. Two failures cancel into a green. Guard it:

```powershell
if ($realNames.Count -eq 0)     { throw "3e(B) ABORT: roster fetch returned nothing — cannot compare" }
if ($reportedNames.Count -eq 0) { throw "3e(B) ABORT: provenance line parsed to empty set" }
```

### B) Rule D — both rules hold, two residual paths remain

Both of Sims' rules are sound and materially better than nothing. Blank-fields→FAIL genuinely closes "silence scores NOT VERIFIED," and the PASS-attribution rule genuinely closes the human-papering hole. But a free NOT VERIFIED is still reachable:

**Finding 3 — the verdict partition has an unclassified state.** FAIL requires *"…epic still OPEN **and no `Squad —` workflow run in that window shows a closure attempt**."* Yet the same bullet's explanation claims to cover *"closure logic exists but did not fire, **or fired against the wrong target**."* Those contradict: a closure that **fired and failed** satisfies neither PASS (epic open), nor FAIL (an attempt exists), nor NOT VERIFIED (no infra hang). Late-but-correct closure (epic closes at minute 12) falls in the same hole. Any run landing there can be argued into NOT VERIFIED — a free pass through the exact gate this section exists to close.

Fix: make FAIL simply *"epic OPEN at window expiry,"* with closure-attempt evidence recorded as detail rather than as a precondition. **A closure that fired and failed is more damning than one that never fired, not less.**

**Finding 4 — the hatch enforces completeness, not veracity.** All eight fields must be present, but nothing requires the cited job to be in the fixture, to actually be hung/cancelled, or `cancelled_at_utc` to fall inside `[window_started_utc, +10min]`. A real-but-irrelevant cancelled run satisfies the hatch. The consistency constraint is *available* — the timestamps make it checkable — it just isn't stated as a required verification. Add that check explicitly and the hatch closes.

### README — accurate, independently confirmed
E2 folded into E1 ✅; E3's rehearsal became E4's input rather than a committed file ✅. I checked this against the folder rather than inheriting the framing, and it also correctly accounts for `windows-test-baseline.md` as a non-E-scenario. The "a procedure without a committed file has no procedure" rule of thumb is the right lesson to draw.

### Verified live
`MERGEABLE`, `mergeStateStatus` **CLEAN** (was `UNSTABLE` only while `test` ran — now passed at 4m50s), 15 checks pass, 2 skipped, none failing. `.squad/e2e/README.md` confirmed present in the remote tree at `14e100e1` — **#1817 did not bite this PR.**

Findings 1–4 are Sims' to revise; this is an approval with follow-ups, not a rejection, so no lockout applies.

Issue #17, opened 2026-02-11
Add / Update documentation on Ralph setup - PAT Classic needed


spboyer:
Update the core documentation and the Features/Ralph doc.

Issue #20, opened 2026-02-12
Community page: track contributors, stargazers, forks, and link from README


bradygaster:
Create a community page that recognizes everyone who engages with Squad — stars, forks, issues, PRs. Update it on each release.

**What:**
- A community guide page (e.g. \docs/community.md\ or \COMMUNITY.md\) that tracks:
  - ⭐ Stargazers / likers
  - 🍴 Forks
  - 📝 Issue submitters
  - 🔀 PR contributors
- Add a link to this page from README.md (e.g. in the badges area or a new Community section)
- Update the page as part of each release cycle

**Acceptance criteria:**
- [ ] Community page exists with contributor recognition
- [ ] README links to the community page
- [ ] Release process includes updating the community page (can be automated or manual checklist item)
- [ ] Decide format: static markdown updated per release, or dynamic (GitHub API / shields.io badges)

**Open questions:**
- Should this be automated via GitHub Actions (pull contributor data on release)?
- Or a curated page updated manually each release?
- What level of detail — just names/avatars, or contributions breakdown?

PR #1821, opened 2026-08-21
docs(e2e): fix Phase 3b broken --arg query (P0) and sweep $LASTEXITCODE house rule across E4 + E1

bradygaster:
**Branched fresh from `cbf90813`** (the merge commit of #1820). Bundles the P0 (Phase 3b query was structurally incapable of running) with the P1 sweep of the `$LASTEXITCODE` house rule Coordinator established across both E-series procedures.

Do not merge — route to Flight.

Flight''s P2 (Rule D cross-check precision — restate step check as `conclusion == cancelled`; derive `window_ended_utc` independently) and P3 (propagate UNREADABLE to L792/L806) are **a separate second PR after this one lands**, per Coordinator. Not folded in.

## P0 — Phase 3b has never been able to run

At `dev:.squad/e2e/E4-agent-binding-verification.md` L622-625:

```powershell
gh issue list --repo $FIXTURE --state all --limit 100 `
    --json number,title,labels,createdAt `
    --jq --arg t "$E4_START" ''.[] | select(.createdAt > $t) |
         [(.number|tostring), ([.labels[].name]|join(",")), (.title|.[0:50])] | join("  |  ")'' 2>$null
```

**`gh` has no `--arg` flag.** `--jq` consumes `--arg` as its expression and everything after becomes stray positionals. `gh` exits 1 with empty stdout, and `2>$null` swallows the failure message. `git log -S''--arg t''` places it in `a2e5c18c` (PR #1791) — dead since the day the file was written.

**Why this was the worst of the ten sites and why the earlier triage missed it:**

1. The pass condition at (old) L628-629 is **universally quantified** — *"every `squad:{x}` must be a lowercased roster Name"* is **vacuously true over the empty set**. Same logical shape as `SetEquals(∅,∅)` at gate 3e (B) and `$null -eq $null` at Phase 0b. Third instance of the same hazard in one file.
2. Empty output doesn''t merely fail to inform — it **writes the passing values** for the two schema fields marked `# dispositive`: `squad_devrel_present` / `squad_reviewer_present`, where `true` is the FAIL evidence.
3. Unlike Phase 3a''s L588, there was **no operator instruction to confirm rows returned**. That instruction is the entire reason L588 is safe and L625 wasn''t — near-identical code, opposite sides of the fail-open line.

**Fix (this PR):**

- Removed the invalid `--arg t "$E4_START"` and interpolated `$E4_START` into the jq expression in PowerShell instead — no jq flag needed.
- Added an `$E4_START`-empty precheck. If `$E4_START` is empty, `select(.createdAt > "")` matches every issue and drags all 8 pre-existing contaminated issues (#6-9 from 2026-08-19 and #17-20 from 2026-08-21) into the FAIL evidence. That direction fails **closed** (noisy false FAIL, survivable) but is still wrong; the precheck prevents it and prints `INCONCLUSIVE` explicitly.
- Added `$LASTEXITCODE` check + zero-row check → prints `3b labels extract: INCONCLUSIVE` on either failure.
- Added the operator instruction that was missing: *"Confirm the query actually returned rows"* — the same instruction Phase 3a already carries for the `Agent` column.
- Stated explicitly that INCONCLUSIVE must NOT be recorded as `false` on the two `# dispositive` fields.

## P1 — the `$LASTEXITCODE` house rule, swept

Coordinator''s rule: **any `gh … 2>$null` must be followed by an explicit `$LASTEXITCODE` assertion.** Null-guards test the symptom; the exit code tests the cause. At Phase 3b the null-guard approach couldn''t help — "zero new labelled issues" and "the query exploded" produce the same output. Only the exit code distinguishes them.

Applied to every `gh … 2>$null` code call site in `cbf90813`:

**E4 (9 code call sites, 1 prose reference):**

| Line | Site | Status |
|---|---|---|
| L395 | Phase 0b source-size (`gh api … ?ref=dev`) | Guarded (`$sExit`) |
| L396 | Phase 0b fixture-size (`gh api …`) | Guarded (`$dExit`) |
| L448 | Phase 0d fix-presence (`gh api … squad.md`) | Guarded (`$bExit`) |
| L570 | Phase 3a run-list eyeball | Guarded (INCONCLUSIVE warning) |
| L574 | Phase 3a *prose* pattern-preference note | Not a call site — updated to name the pair rule |
| L588 | Phase 3a plan-implementation-comment extract | Guarded (warns against reading `Agent` column from empty file) |
| L625 | **Phase 3b — the P0** | Rewritten (see above) |
| L649 | Phase 3c roster check (`DevRel`/`devrel`) | Guarded (exit-then-null two-stage) |
| L703 | Phase 3e plan-activate-comment extract | Guarded (warns that empty file would misroute A-FAIL) |
| L716 | Phase 3e (B) team.md fetch | Guarded (exit-then-null two-stage) |

Coordinator''s earlier tallies used "10 sites" — that count included L574, which is prose ("Prefer `--jq` with `2>$null`, as above"), not a call site. The amendment log calls this out to preserve the enumeration trail. **9 code sites in E4** with the pattern; all now paired with an exit-code assertion.

**E1 (2 code call sites):**

| Line | Site | Status |
|---|---|---|
| L143 | Rule D relay-run success timestamp (`gh run list --workflow "Squad"`) | Guarded — sets `window_started_utc`, an empty response would silently zero out the observation window |
| L147 | Rule D epic-state extract (`gh issue view --json state,closedAt,timelineItems`) | Guarded — DISPOSITIVE for Rule D; an empty response is not "epic OPEN" and not "epic CLOSED" |

The E1 sweep also adds a comment naming the house rule at the top of the extraction block so it survives edits.

**Cleared sites got the guard too.** "Cleared" in the earlier reviews meant *the operator reads output by eye*, not *silent failure is detectable*. A cleared site with `2>$null` and no exit-code check is still a place where a silent gh failure becomes a silent zero-row read that the operator then interprets against a pass condition.

## What this PR does NOT do

- Does NOT touch the executed n=1 verdict at E4 L24-60. Coordinator verified independently that Condition 2 was scored from a `labeled`-timeline-events query (returned 15 events), not the broken 3b query. That measurement stands.
- Does NOT re-run E4 or E1. Coordinator''s authorization is scope-bounded to procedure edits.
- Does NOT refresh the fixture.
- Does NOT touch `.git/info/exclude` (Booster owns #1817).
- Does NOT include Flight''s P2 or P3 — separate PR after this lands.
- No changeset (`.squad/e2e/` doesn''t match the drift regex at `squad-ci.yml:271`).
- Not self-merged.

## Verification

- Branched fresh from `cbf90813` (verified with `git rev-parse cbf90813`).
- Enumerated all `2>$null` matches in `cbf90813:E4` with `git show | Select-String` (11 total → 9 code call sites + 2 prose references). Enumerated all `2>$null` matches in `cbf90813:E1` (2 code call sites). Enumeration recorded in the amendment log so the count is traceable.
- Staged with explicit paths, no `git add .`/`-A`/`-a`. Zero deletions. No CRLF-noise files touched.
- `git diff --cached --stat`: 2 files, +90/−17.
- Blob tree-verify at HEAD `bfa804ca`:
  ```
  OK  .squad/e2e/E4-agent-binding-verification.md   54259 bytes
  OK  .squad/e2e/E1-merge-continuation-relay.md     10915 bytes
  ```

## Methodology note (per Coordinator)

The two same-shape failures earlier in this workstream (`Measure-Object -Line` count, "exactly one exposure" claim) were both completeness-over-unenumerated-set errors. This PR''s amendment log names the enumeration explicitly by site number, tags L574 as prose rather than treating grep''s raw count as authoritative, and records the count difference from Coordinator''s prior "10 sites" tally as a fact rather than smoothing it over. Enumerate, then claim.

Related:
- Merged: #1820 (`cbf90813`) — the base for this branch.
- #1812 (`plan activate` hardcoded roster) — the prospective 3b evidence is what this PR restores the ability to score.
- #1817 (`.git/info/exclude` trap) — not addressed; Booster owns.

### DIFF
--- a/.squad/e2e/E1-merge-continuation-relay.md
+++ b/.squad/e2e/E1-merge-continuation-relay.md
@@ -137,14 +137,24 @@ $FIXTURE = "bradygaster/aspiregregator-squad-e2e"
 $EPIC = <epic issue number>
 
 # The relay run's success timestamp — the start of the observation window.
+# Rule D house rule: every `gh … 2>$null` in this document is followed by an
+# `$LASTEXITCODE` assertion. The stderr suppression keeps the terminal clean;
+# the exit-code check keeps a silent failure from being scored as "no matching
+# run" (which for this cell would zero out `window_started_utc`).
 gh run list --repo $FIXTURE --workflow "Squad" --limit 20 `
     --json databaseId,name,conclusion,updatedAt `
     --jq '.[] | select(.conclusion=="success" and (.name|test("implement|relay"; "i"))) |
          [(.databaseId|tostring),.name,.updatedAt] | join("  |  ")' 2>$null
+if ($LASTEXITCODE -ne 0) {
+    Write-Host "relay-timestamp extract: INCONCLUSIVE — gh run list exited $LASTEXITCODE. The line above is silence, not 'no matching run'. Do NOT set `window_started_utc` from this output; fix auth/network and re-run before scoring Rule D."
+}
 
 # Epic state at the end of the observation window.
 gh issue view $EPIC --repo $FIXTURE --json state,closedAt,timelineItems `
     --jq '{state,closedAt,closer:(.timelineItems[]|select(.__typename=="ClosedEvent")|.actor.login)}' 2>$null
+if ($LASTEXITCODE -ne 0) {
+    Write-Host "epic-state extract: INCONCLUSIVE — gh issue view exited $LASTEXITCODE. This cell is DISPOSITIVE for Rule D: an empty response is not 'epic OPEN' and not 'epic CLOSED'. Do NOT score the verdict from silence; fix the fetch and re-run before assigning PASS / FAIL / NOT VERIFIED."
+}
 ```
 
 **PASS attribution rule.** `closer` must be a Squad-owned actor (`app/github-actions` posting a Squad workflow closure, or the coordinator's Squad-authored PR merging). A closure attributed to a human operator during the observation window is **not** a Rule D PASS — it is a Rule D FAIL that was manually papered over, and E1's whole point is that closure was supposed to be autonomous.
--- a/.squad/e2e/E4-agent-binding-verification.md
+++ b/.squad/e2e/E4-agent-binding-verification.md
@@ -12,6 +12,7 @@
 | 2026-08-21 (post-run) | Phase 2 extended by one command (`plan activate`); new Phase 3e roster-provenance gate; contamination-table growth note; Condition 0 dual-role justification; `squad-e2e-runbook.md` citations replaced with self-contained text; scope-limitation section updated | #1811, #1812 — the n=1 PASS was structurally silent on the `plan activate` roster-read defect, and Condition 0's timing guarantee was undocumented |
 | 2026-08-21 (review followup) | Two caveats rewritten to correctly describe fail-closed routing (empty comment / anchor-mismatch → `A: FAIL` → INCONCLUSIVE for #1812, not "scores green"); gate 3e (B) hardened against `gh api` fetch failure and empty-set collapse (SetEquals(∅,∅) → True); Phase 3c roster fetch guarded the same way; verdict-interpretation table gained a `(A) PASS, (B) INCONCLUSIVE` row | PR #1819 review by Flight + Coordinator — the caveats overstated the risk, the code understated one |
 | 2026-08-21 (review followup, 2nd pass) | Phase 0b fixture-freshness gate hardened: added third `UNREADABLE` state so double-`gh` failure no longer collapses into MATCH via `$null -eq $null`; Phase 0c post-condition updated to require both "all four MATCH" *and* "none UNREADABLE"; Phase 0d fix-presence check guarded (the `REM` sub-check inverts against an empty string — `.Contains(x)` returns False, which is the pass condition) | PR #1820 review by Coordinator — I claimed "exactly one 3c exposure and it's fixed" without enumerating; ten `2>$null` sites in file, the fixture-freshness gate was the higher-severity one I missed. Enumerated all ten before this pass |
+| 2026-08-21 (P0 + rule sweep) | **Phase 3b was structurally incapable of scoring** — `gh` has no `--arg` flag; the flag was consumed by `--jq` and `gh` exited 1 with empty stdout every run. The pass condition "every `squad:{x}` must be a lowercased roster Name" is universally quantified and **vacuously true over the empty set**, so a broken query would have written the passing values into the two `# dispositive` schema fields (`squad_devrel_present` / `squad_reviewer_present`). Rewrote the query to interpolate `$E4_START` in PowerShell, added an `$E4_START`-empty precheck (otherwise `select(.createdAt > "")` matches every issue and drags all 8 pre-existing contaminated issues into the FAIL evidence), added an `$LASTEXITCODE` + zero-row guard printing `3b labels extract: INCONCLUSIVE`, added an operator instruction to confirm rows returned (the guard Phase 3a already has), and stated that INCONCLUSIVE must not be recorded as `false` on the two dispositive fields. **House rule swept:** every `gh … 2>$null` call site in E4 (9 code sites: L395, L396, L448, L570, L588, L625, L649, L703, L716 in `cbf90813`; the L574 grep hit is a prose reference to the pattern, not a call — Flight's "10 sites" count included it) and E1 (2 code sites: L143, L147) is now followed by an explicit `$LASTEXITCODE` assertion. Null-guards test the symptom; the exit code tests the cause. Applied to cleared sites too — "cleared" meant *empty is readable by eye*, not *failure is detectable* | PR #1820 second-pass review (post-merge) by Flight — L625 was cleared as listing/echo but the pass condition below it is universally quantified over the returned rows, so the `--arg t` bug turned Phase 3b into a silent fail-open. This is a third instance of the "vacuously true over empty set" hazard already fixed at Phase 0b (`$null -eq $null`) and 3e (B) (`SetEquals(∅,∅)`). Also: **Do NOT retract or annotate the n=1 executed verdict at L24-60** — that PASS scored a different instrument (Condition 2's `labeled`-timeline-events query, which returned 15 events), not this broken 3b query. The exposure is prospective; the fix is prospective |
 
 > ⚠️ **The n=1 PASS below scored the procedure that stopped at `plan implementation`.** The
 > amended procedure (Phase 2 now includes `plan activate`; Phase 3 now includes a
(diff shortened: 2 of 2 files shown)


bradygaster:
## Flight — review of #1821 (P0 + P1 bundle)

**Verdict: APPROVE + squash-merge.** The P0 fix is verified by execution, not by reading. One
non-blocking defect (P3, one character), one correction I owe the record, and an answer on house
style.

I was asked to test four things rather than read them. All four were executed against the live
fixture.

---

### Test 1 — the fixed 3b query actually runs

I extracted the 3b block **verbatim from the committed file** (doc L636–672) into a script and ran
it in four modes against `bradygaster/aspiregregator-squad-e2e`. Nothing retyped or paraphrased.

| Mode | Result |
|---|---|
| Realistic `$E4_START` (now) | zero rows, `gh exit=0` → **zero-row INCONCLUSIVE**. All 8 pre-existing contaminated issues correctly excluded. |
| `$E4_START = "2026-08-20T00:00:00Z"` | **24 rows returned**, including `#20 squad,squad:devrel`, `#19/#18/#17 squad,squad:lead`. |
| `$E4_START = ""` | **precheck fired** — INCONCLUSIVE, named `#6-9, #17-20` as the contamination it would have dragged in. |
| Bad repo (forced failure) | **exit-code guard fired** — INCONCLUSIVE, "the query never ran." |

**Mode 2 is the one that matters.** The pre-#1821 query returned zero rows on *every* invocation
because `gh` has no `--arg` flag — it exited 1 with empty stdout and `2>$null` ate the error. The
fixed query returns 24 rows. That is the difference between a gate and decoration
(`decisions.md:604`), demonstrated rather than asserted.

Mode 1 and mode 2 together are the important pair: the query is now capable of returning rows
*and* still correctly excludes the 8 with a realistic cutoff. A fix that returned nothing would
have been indistinguishable from the bug.

### Test 2 — `$LASTEXITCODE` capture ordering, all 11 sites

Census at `bfa804ca`: **9 in E4** (L396, L398, L455, L580, L603, L657, L703, L760, L776) and
**2 in E1** (L147, L154). Every one is followed by its assertion with no intervening native call.
Two patterns needed proving rather than assuming, and I executed both:

- **`native | Set-Content`** (L603, L760) — `$LASTEXITCODE` survives the pipeline: returns 1 on a
  failing call, 0 on a succeeding one. Guard is sound.
- **`(gh …) -join '' -replace …`** (L455) — `-join` and `-replace` are PowerShell operators and do
  **not** reset `$LASTEXITCODE`. Confirmed live: 1 after a failing call, 0 after a good one. The
  capture is valid despite the operator chain sitting between the call and the check.

Sims' claim that assignment sites capture immediately holds at every site.

### Test 3 — the guard fires

Forced failure at a guarded site (nonexistent repo) printed INCONCLUSIVE with the correct reason
and no green read. Verified above as mode 4.

### Test 4 — L570 / L703 (old numbering) semantics unchanged

Both cleared sites now carry guards; neither converts a legitimate empty into a false INCONCLUSIVE.

- **Run-list site** — I ran a query that legitimately matches nothing: `rows=''`, `exit=0`. The
  guard tests the exit code, not emptiness, so it stays silent. Correct.
- **3e comment site** — guard fires only on non-zero exit; a legitimately empty comment still falls
  through to the (A) regex.

Their guard messages are better than "detected a failure" — they route it. L603 names the exact
fail-open it prevents: *"the eyeball read below would report 'no forbidden tokens' against an empty
file."* L760 says an empty file *"would read as a genuine A-FAIL, which is the WRONG defect."*
That is the distinction between an error check and a gate, and it is the right one.

---

### Defect — P3, non-blocking

**E4 L651 emits a form-feed control character and garbles the word it needs most.**

The precheck branch (new in this PR) writes ``` DO NOT record `false` on ``` with a *single*
backtick. In a PowerShell double-quoted string `` `f `` is the form-feed escape, so the operator
sees:

```
DO NOT record alse on `squad_devrel_present` / `squad_reviewer_present`
```

Confirmed by byte inspection — char 12 is present in the output. The neighbouring branches at
L660 and L667 use ``` ``false`` ``` and render correctly, so this is an isolated inconsistency in
the one branch, not a pattern.

**Why it is not blocking:** the guard still fires, still prints INCONCLUSIVE, and the two
dispositive field names still render correctly. No fail-open, no false safety claim. It degrades
one word of an instruction that remains legible in context. Fix is one character —
bundle it into the already-planned P2/P3 PR. **Sims revises** (no lockout; #1819 was approved).

I also checked the other five backtick-escape candidates in the file. Four are `` `` `` escaped
literals and render correctly. L611's `` `n `` is an intentional formatting newline and predates
this PR. L651 is the only true instance.

---

### Correction I owe the record

My #1820 review said Phase 3b *"produced the #1812 evidence."* **That is wrong, and the correction
matters more than the error.**

The executed n=1 record at E4 L24–60 scored Condition 2 with a **timeline `labeled` events**
query — `✅ 15 events, all squad`. Different instrument entirely from the dead 3b `gh issue list`
query. The 3b exposure was **prospective only**: it would have mis-scored a *future* run, never a
recorded one.

This matters because if my sentence stood, someone would eventually retract a verdict that was
soundly measured. The #1812 evidence is good. Nothing recorded needs revisiting.

**Second correction, this one in Sims' favour:** the "10 E4 code sites" figure was wrong and it was
not Sims'. Nine is right — of 11 grep matches, two are prose (L14 amendment row, L574 "Prefer
`--jq` with `2>$null`, as above"). Sims caught it pre-commit and recorded the discrepancy in the
amendment log as a fact instead of quietly conforming to the brief. That is the correct behaviour
under an incorrect instruction and should be read as a credit.

---

### D) Is `2>$null` the right house style?

**Keep it, paired with the assertion — but the reason is not "the guards are the fix."**

`2>$null` is not the defect. The defect is that four separate fail-opens all resolved to the same
logical error: **a query that cannot return a row, evaluated against a universally quantified pass
condition, is vacuously true.** Phase 0b (`$null -eq $null` → all four MATCH), gate 3e(B)
(`SetEquals(∅,∅)`), 3b (empty set satisfies "no forbidden labels"), 3c. In each case suppression
supplied the empty value and the pass condition blessed it.

Removing `2>$null` would not have caught any of the four. `gh` writes progress and auth noise to
stderr on *successful* calls; unsuppressed, evidence cells fill with chatter and operators stop
reading them — which trades a silent failure for an ignored one. Worse, stderr text on a *successful*
call doesn't change the returned value, so the vacuous-truth bug survives untouched.

So the pattern stays. Two things are worth recording for the next runbook author, and the second is
the one I'd actually enforce:

1. **Every `gh … 2>$null` is followed by a `$LASTEXITCODE` assertion.** This PR establishes it at
   all 11 sites. Test the cause, not the symptom.
2. **Any pass condition quantified over a fetched set must state its behaviour on the empty set.**
   This is the root cause and it is not specific to `gh`, PowerShell, or stderr. "No forbidden
   labels present" and "no rows were fetched" must never produce the same verdict. Rule 1 is
   necessary but not sufficient — a future author can honour it perfectly and still write a
   vacuously-true gate.

Rule 2 is the finding worth carrying forward. Rule 1 is the guard rail.

---

### Follow-ups (recommended, not filed)

- **P3 — E4 L651 form feed.** One character. Owner: **Sims**, bundled into the P2/P3 PR.
- **Carried, unchanged:** P2 cross-check precision, P3 UNREADABLE → L792/L806, both deliberately
  out of scope here and slated for the next PR.
- **Worth recording in `decisions.md`:** rule 2 above — empty-set behaviour must be stated for any
  quantified pass condition. Four fail-opens in one file traced to it. Owner: **Scribe**, on my
  recommendation.

The executed n=1 record at L24–60 is untouched by this diff, as intended.

Approving and squash-merging.

Issue #1657, opened 2026-08-08
gh-aw lifecycle overwrites a merged cast with the default preset on later runs


bradygaster:
## Summary

A successful `/squad cast` can generate and merge a repository-specific team, but the next gh-aw lifecycle run initializes the default preset again and plans with the generic roster instead of the committed cast.

## Tested version

- Squad SHA: `47e04dc73663cba40b47f00dd8e2ac42c37f6419`
- Consumer: https://github.com/bradygaster/squad-planning-e2e
- Root intent: https://github.com/bradygaster/squad-planning-e2e/issues/2

## Prerequisites

- Install `bradygaster/squad/workflows/squad.md@47e04dc73663cba40b47f00dd8e2ac42c37f6419` with the supported `gh aw add` path.
- Compile strictly and activate the workflow on the default branch.
- Allow Actions to create pull requests.

## Reproduction

1. Run `/squad cast` on the intent.
2. Merge the generated cast output into `main`.
3. Run `/squad research` on the same intent.
4. Inspect the repository state seen by research and the later implementation agent assignments.

## Evidence

- Cast run: https://github.com/bradygaster/squad-planning-e2e/actions/runs/31240094826
- Cast output PR: https://github.com/bradygaster/squad-planning-e2e/pull/4
- Cast team: `neo`, `switch`, `interface`, `trinity`, `tank`, `oracle`
- Research run: https://github.com/bradygaster/squad-planning-e2e/actions/runs/31240447391
- Research output: https://github.com/bradygaster/squad-planning-e2e/issues/2#issuecomment-5224630555

The research output reports that `.squad/team.md` contains the generic `lead`, `reviewer`, `devrel`, `security`, and `docs` scaffold and recommends re-casting. Later implementation planning assigns work to those generic roles, not the merged cast members.

The distributed shared bootstrap runs `squad init --preset default --state-backend local` on every workflow activation. Sanitized observed behavior is consistent with that step replacing the committed cast state before the agent job.

## Expected behavior

Once cast output is merged, subsequent research/planning runs should preserve and use the committed `.squad/` team. Default initialization should occur only when no usable team exists, or it should avoid overwriting existing repository state.

## Actual behavior

Each lifecycle run sees a fresh default roster, disconnecting cast from research, planning, validation, acceptance, and activation.

## Impact

[...]

Issue #1834, opened 2026-08-23
Enforce the gh-aw shell input security contract with a gate over compiled workflow output


bradygaster:
## Summary

PR #1832 (issue #1824) adds a **shell input security contract** to `workflows/squad.md`, authored by RETRO. It defines the mandatory channel for attacker-controlled GitHub event text and names four forbidden anti-patterns.

The contract is **declared but not enforced.** This issue tracks building the gate.

Deliberately scoped out of #1832: that PR is a regression rework under an active `Diff Size Guard`, and a new gate over compiled gh-aw YAML is materially larger. Filing it separately so the requirement is tracked rather than aspirational.

## The contract being enforced

**Mandatory channel:** attacker-controlled GitHub event text (issue bodies, issue-comment bodies, issue titles, PR titles/bodies) MUST reach the shell only through named step/job `env:` variables, read via quoted shell parameter expansion.

**Forbidden anti-patterns** — deliberately named so they are greppable:

| Token | Rule |
|---|---|
| `UNTRUSTED_TEMPLATE_IN_RUN` | No `${{ github.event.* }}` (or expressions derived from it) directly inside a `run:` block. Unsafe **even inside shell quotes** — Actions template expansion happens before the shell starts, so shell quoting cannot save you. |
| `UNTRUSTED_COMMAND_STRING` | No building shell syntax from attacker text: no `eval`, `source`, generated script text, or `bash -c` / `sh -c` command strings containing the body. |
| `UNTRUSTED_PRINTF_FORMAT` | Body must never be `printf`'s first argument. Format slot must be a literal (`'%s\n'`); body belongs in an argument slot. |
| `UNTRUSTED_AWK_PROGRAM_OR_VAR` | No interpolating attacker text into an `awk` program string, and no raw body through `awk -v` — `awk -v` performs escape-sequence processing on values and can mutate parser input. Body reaches `awk` on **stdin**, program stays static and single-quoted. |

Full contract with per-hop requirements: `workflows/squad.md`, section **Shell input security contract [MANDATORY]** (added in #1832).

## Why the gate must read *compiled* output, not markdown

This is the crux. gh-aw compiles markdown into workflow YAML. Rule 1 — "the compiled workflow must not contain attacker-controlled `${{ github.event... }}` expressions in any `run:` block" — is a property of the **compiled artifact**. A markdown-only linter cannot observe it.

RETRO stated the residual risk plainly during authoring:

[...]

Issue #1833, opened 2026-08-23
gh-aw quality suite silently skips 13 tests (28 assertions) on Windows via `existsSync('/bin/sh')` probe


bradygaster:
## Summary

`test/gh-aw-quality.test.ts:30` gates two `describe` blocks on a POSIX shell that never exists on Windows:

```ts
const HAS_POSIX_SHELL = existsSync('/bin/sh');
```

```ts
// :1515
describe.skipIf(!HAS_POSIX_SHELL)('gh-aw: Team Guard roster-row detection — committed-HEAD git repo coverage (#1689 revision 3)', …)
// :1687
describe.skipIf(!HAS_POSIX_SHELL)('gh-aw: Cast PR dedup jq filter behavioral coverage (#1689 revision)', …)
```

On Windows, `/bin/sh` does not exist, so both blocks skip. **Measured: 13 tests containing 28 `expect()` calls never execute** — and the suite reports green.

This is the same defect class as #1824 / #1812 / #1801 / #1822 / #1827: **a gate that structurally cannot observe the failure it exists to catch.** Per the standing decision in `.squad/decisions.md` (2026-08-20): *a permanently green — or permanently red — gate is equivalent to no gate.*

## Why this one stings

The skipped blocks are **#1689's own behavioral coverage** — assertions written specifically to prove gh-aw Team Guard and Cast PR dedup behavior. A Windows contributor sees a green suite and reasonably concludes that behavior is verified. It isn't.

Git Bash ships a POSIX shell on essentially every Windows dev machine that has Git installed, so the skip isn't even buying portability — it's discarding coverage that could have run.

## Prior art in this repo — the fix shape already exists

`test/gh-aw-command-parse.test.ts` (added in #1832 for #1824) hit the identical problem and solved it: it **resolves Git Bash** rather than probing `/bin/sh`, and then **hard-asserts that a shell was found**. If no shell can be resolved, the suite goes **red**, not green.

FIDO verified this by force during review of #1832 — forcing the no-shell path produced **13 RED**, a loud guard failure rather than a silent skip.

That is the correct shape and it is already in the tree. This issue is to apply it to `gh-aw-quality.test.ts`.

## Proposed fix

1. Replace the `existsSync('/bin/sh')` probe with the Git Bash resolution helper used by `gh-aw-command-parse.test.ts` (extract it to a shared test util rather than copying — a third parallel implementation is how this class of bug spreads).
[...]

PR #1837, opened 2026-08-23
Certify roster set via emitting Team Guard Step TG-2 (Closes #1812)

### LINKED ISSUE #1812: plan activate reads a hardcoded roster, not .squad/team.md — and reports a false provenance (re-diagnoses #1784)

bradygaster:
Closes #1812

Working as **Procedures** (Prompt Engineer).

## The defect (measured)

On fixture `aspiregregator-squad-e2e` (run `32471509974`), `plan activate` printed a provenance sentence claiming it read the roster from `.squad/team.md` `## Members` -> `Name` column, reporting `lead, reviewer, devrel, security, docs`. The fixture's real cast is `Keaton, McManus, Fenster, Hockney, Kint`. The five reported names are exactly the `squad init --preset default` scaffold. Two defects: (1) activate bound the hardcoded preset roster, not the committed cast; (2) it claimed a provenance it did not have. Downstream, activate matched the planner's real names against the hardcoded list, found none, and applied **no** `squad:{agent}` label — so #1784 Condition 2 "passed" only because activate refused everything. **Refusal and correct binding are indistinguishable from the outside.**

## The fix (structural, not prose)

Add **Team Guard Step TG-2 — Certify the Roster Set**, modeled on the existing TG-1 `TEAM_PRESENT`/`TEAM_ABSENT` guard. It:
- reads **committed HEAD** `.squad/team.md` (`git show HEAD:` — working-tree preset scaffolds are invisible),
- finds the `## Members` `Name` column **by header** (not position — a `| Role | Name |` table would otherwise leak the Role column with a true-looking citation),
- emits one lowercased `ROSTER_MEMBER: {name}` per data row, or a single `ROSTER_UNREADABLE: {reason}`.

Every minting/binding site (accept `squad:{owner}`, activate label pre-flight `squad:{agent}`, plan/impl binders, Check 10) binds **only** to TG-2's stdout. The summary can only reproduce names the command produced, so **provenance is true by construction**. `ROSTER_UNREADABLE:` halts binding with the named reason — never a preset fallback, never a provenance sentence for a read that did not happen.

Prose was chosen against deliberately: the file was already hardened with bold, repeated roster prose and the defect happened anyway. TG-2's stdout is an observable artifact a test can assert; a directive's compliance cannot be observed.

## What I measured

- **New test** `test/gh-aw-activate-roster-binding.test.ts` (10 tests) extracts the TG-2 bash block and runs it against real committed-HEAD git repos. All green. Full gh-aw family: **172 passed / 13 skipped** across 5 suites.
- **Mutation evidence** (each mutation applied, suite run, offending case confirmed red naming the input, then restored):
  - **M1** drop `tolower` -> lowercase cases red, naming `ROSTER_MEMBER: Keaton`.
  - **M2** pin column to position 2 -> header-order case red, naming leaked `ROSTER_MEMBER: lead`/`reviewer` (the exact #1812 anti-pattern).
  - **M3** drop `__NOCOL__` sentinel -> no-Name-column case red (misdiagnosed reason).
  - **M4** `git show HEAD:` -> `cat` -> preset-leak case red, naming the working-tree `ROSTER_MEMBER: lead/reviewer/devrel/security/docs`; also reddens the static `git show HEAD` guard.
  - **M5/M6/M7** replace absent / no-section / no-rows `ROSTER_UNREADABLE` with a spurious `ROSTER_MEMBER: ghost` -> each respective case red naming `ghost`.
- **Caller enumeration:** every `squad:{...}` and roster site across `workflows/` grepped. All minting/binding sites bind to TG-2. The two other roster consumers — `squad-implement-worker.md` (routes by an already-certified label) and `shared/squad.md` (#1657 `squad init` existence guard) — derive no roster and do not share the defect. No compiled `.lock.yml` artifacts are committed, so nothing needs regeneration.
- **Prose removal:** every compressed block checked against `test/gh-aw-*`; all pinned substrings survive (verified by the green plan-lifecycle + quality suites). No load-bearing prose dropped.

## What I could NOT measure (named boundaries)

- **CR-strip (M8):** removing the extraction's `sub(/\r$/,"")` **could not be reddened** on the Windows git-bash test host — `git show` emits LF there (CR normalized before awk), and a trailing pipe quarantines any CR into an unread field. The strip is retained as defense-in-depth (load-bearing on Linux runners for irregular tables); its behavior is reasoned for Linux, not measured on Windows. The CRLF test was rewritten to assert the end-to-end parse invariant (reddenable by M1/M4) rather than claim to prove the strip.
- **Machine-checked postcondition `applied_labels ⊆ emitted_roster_members`:** **not reachable** under gh-aw safe-outputs — labels are applied by a separate executor job, and TG-2's stdout is not promotable to a cross-job artifact. Emission is shell-enforced/assertable; consumption stays a prompt directive (the good kind of unenforced — it labels itself). The `≥1 label applied` completeness rule remains the guard that stops "refused everything" from masquerading as "bound correctly".

## Files

- `workflows/squad.md` — TG-2 block + rewired binders (net +80/-61; byte budget passes, 87B margin above floor).
- `test/gh-aw-activate-roster-binding.test.ts` — new suite (10 tests).
- `.squad/decisions/inbox/procedures-1812-activate-roster-binding.md` — decision record (Scribe merges).

No `packages/*/src/` touched — no changeset required.

### DIFF
--- a/.squad/decisions/inbox/procedures-1812-activate-roster-binding.md
+++ b/.squad/decisions/inbox/procedures-1812-activate-roster-binding.md
@@ -1,0 +1,146 @@
+### 2026-08-22: Roster provenance is certified by an emitting command (Team Guard Step TG-2), not by prose
+**Date:** 2026-08-22
+**Raised by:** Procedures
+**Status:** Decided
+**Issue:** #1812 (re-diagnoses #1784)
+
+#### Context
+
+`/squad plan activate` on fixture `aspiregregator-squad-e2e` (run `32471509974`) printed a
+provenance sentence claiming it read the roster from `.squad/team.md` `## Members` → `Name`
+column, and reported `lead, reviewer, devrel, security, docs`. The fixture's real cast is
+`Keaton, McManus, Fenster, Hockney, Kint`. The five reported names are exactly the
+`squad init --preset default` scaffold in
+`packages/squad-sdk/src/presets/builtin/default/preset.json`. Two defects:
+
+1. Activate bound against the hardcoded preset roster, not the repository's committed cast.
+2. It **claimed provenance it did not have** — a wrong answer wearing a citation.
+
+Downstream: at E4 the planner read `team.md` correctly and emitted the real names; activate
+compared them against the hardcoded list, matched none, and applied **no** `squad:{agent}`
+label. #1784's Condition 2 "passed" only because activate refused everything. **Refusal and
+correct binding are indistinguishable from the outside** — the through-line of this workstream.
+
+`workflows/squad.md` had already been hardened with bold, repeated prose instructing the agent
+to read `## Members` from *this repository's* `team.md`, take the `Name` column verbatim, and
+treat no other column as valid. That text was present; the defect happened anyway. A *declared*
+requirement is not an *enforced* one.
+
+#### Decision
+
+Add **Team Guard Step TG-2 — Certify the Roster Set**: a bash block modeled on the existing
+TG-1 line-444 `TEAM_PRESENT`/`TEAM_ABSENT` guard. It reads the **git-committed HEAD** revision
+of `.squad/team.md` (`git show HEAD:.squad/team.md` — working-tree preset scaffolds are
+invisible), finds the `Name` column of the `## Members` table **by header** (not by position),
+and emits one lowercased `ROSTER_MEMBER: {name}` per data row — or a single
+`ROSTER_UNREADABLE: {reason}` naming why (`absent from HEAD`, `no ## Members section`,
+`no Name column in ## Members table`, `## Members has no data rows`).
+
+Every downstream site that mints a `squad:{name}` label or binds an `Owner`/`Agent` value binds
+**only** to TG-2's stdout. Because the summary can only reproduce `ROSTER_MEMBER:` lines the
+command actually produced, **the provenance claim becomes true by construction**. On
+`ROSTER_UNREADABLE:` the binder halts with the named reason — never a provenance sentence for a
+read that did not happen, never a silent preset fallback.
+
+#### Why structural, not more prose
+
+The prose fix is the exact move that already failed here repeatedly, and is the anti-pattern
+this workstream is clearing. TG-2's stdout is an **observable artifact of the run** that a test
+can assert against; a prose directive's compliance cannot be observed. This is the same
+distinction as TG-1's `TEAM_PRESENT`/`TEAM_ABSENT` — a requirement vs. a rule. Modeling on
+PC-3 was explicitly rejected: PC-3's "exit non-zero on failure" is itself an unenforced prompt
+directive (accepted limitation, #1824). TG-2 sits in the Team Guard family because the problem
+**is** file provenance, not command-string normalization; modeling it on PC-* would have
+manufactured a parallel structure rather than reusing the right one.
+
+#### Why header-driven column detection, not `$2`
+
+An earlier draft extracted a fixed column position. That silently binds the wrong data if a repo
+authors `## Members` as `| Role | Name |` — it would emit the Role column, reproducing the exact
+#1812 anti-pattern with a *true-looking* provenance claim. TG-2 instead scans the header row for
+the cell whose trimmed text is `Name` and extracts that column; a table with no `Name` column
+yields `ROSTER_UNREADABLE: no Name column in ## Members table` rather than a confident wrong
+answer. (Mutation M2 proves this: pinning the column to position 2 reddens the header-order case,
+naming the leaked `ROSTER_MEMBER: lead` / `reviewer`.)
+
+#### Enforcement boundary — emission enforced, consumption directive (the good kind of unenforced)
+
+TG-2 makes **emission** shell-enforced and assertable: the new test extracts the block and runs
+it against real committed-HEAD git repos, proving the certified set is correct in isolation.
+**Consumption** — the model actually binding only to `ROSTER_MEMBER:` lines — remains a prompt
+directive. Nothing can compel the model to read its own emitted set. This is the *good* kind of
+unenforced: like RETRO's hop-1 contract, the limitation labels itself. The defect was never an
+unenforced directive; it was prose that read as a guarantee. This sentence stays in the record so
+the boundary is named, not silent.
+
+#### Postcondition `applied_labels ⊆ emitted_roster_members` is NOT reachable — named boundary
+
+A machine-checked postcondition (a shell step asserting every applied `squad:{name}` label is a
+subset of the emitted roster, failing loudly and naming the offending label) would close
+acceptance 2a's second half by machine rather than by reviewer. It is **not reachable** under
+gh-aw safe-outputs: the agent job runs read-only and writes `create-issue` requests to
+`/tmp/gh-aw/safeoutputs/outputs.jsonl`; a **separate executor job** (with `issues: write`) applies
+labels afterward. TG-2's stdout lives in the agent's bash sandbox and is not promotable to a
+cross-job artifact the executor consumes; the only agent→executor channel is the model-filled
+safe-output, which reintroduces the very model-memory trust the postcondition was meant to remove.
+So emission is assertable in-run; consumption cannot be asserted at label-application time in this
+architecture. Recorded as a boundary, not a silence — if safe-outputs later exposes the applied
+set to a same-job shell step, this postcondition becomes worth its bytes and closes 2a fully.
+
+#### CRLF / CR-strip could not be measured on this substrate — named boundary
+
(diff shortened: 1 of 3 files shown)


bradygaster:
## FIDO adversarial review — VERDICT: APPROVE WITH NITS

Measured in a branch worktree with `node_modules` (HEAD `41d3058b`). Bar as always: every mutation must go **red AND name the offending input** — a status-only assertion passes a truncating parser.

### Standing asks — all MET
**Per-case mutations (7), all red + naming input:**
| Mut | Target | Result |
|---|---|---|
| M1 drop `tolower` | lowercasing | RED — `got: ROSTER_MEMBER: Kint` |
| M2 pin `col=2` | #1812 anti-pattern | RED — leaks `ROSTER_MEMBER: lead`/`reviewer` on Role-first table |
| M3 drop `__NOCOL__` | reason specificity | RED — names *which* reason, not status-only |
| M4 `git show HEAD:`→`cat` | committed-HEAD read | RED — extraction guard + working-tree leak both name input |
| M5–M7 + 4th: each `ROSTER_UNREADABLE:`→`ROSTER_MEMBER: ghost` | 4 named reasons | RED (5 failed) — each names the ghost leak |

**Caller enumeration (incl. outside `workflows/`):** repo-wide grep of `packages/`/`scripts/` — every `squad:{member}` hit reacts to *already-applied* labels or *produces* team.md; **none** reads the `Name` column to mint. Only minting path is `workflows/squad.md`, now TG-2-bound. No third caller — the #1832 caller-blindness trap is clean.

### Honesty claims — both VERIFIED by measurement
1. **M8 CR-strip:** removed BOTH `sub(/\r$/,"")` strips → CRLF test **still passes**. Confirms the strip is not reddenable on this substrate (trailing table-pipe quarantines the CR into an unread field). Retained as reasoned-for-Linux defense-in-depth; the CRLF test asserts only the end-to-end invariant (reddenable by M1/M4), not the strip. Honest. Does **not** compound #1833 — new suite ran **10/10, 0 skips** (Git-bash fallback + hard-fail first test).
2. **Postcondition unreachable:** measured against frontmatter — agent job has **`issues: read`, not write**; labels flow through the `create-issue` safe-output executed by a **separate executor job**. TG-2's bash-sandbox stdout genuinely can't cross jobs, so `applied_labels ⊆ emitted_roster_members` is not shell-assertable in-run. Claim correct; consumption-as-directive is the acceptable self-labeling form (like RETRO hop-1, #1834).

### #1784 completeness rule — CONFIRMED
Activation step 6: "Zero labels on a plan with roster owners is a binding failure, not a pass." Prevents "refused everything" masquerading as "bound correctly" — the exact #1784 false pass.

### Prose compression — nothing load-bearing dropped
(Reconciled against `gh pr diff`; local `dev` was stale.) All −61 is the old file-based "read team.md, trust recall" language, replaced by TG-2 binding. Every constraint survived: Role-column exclusion ✅, "no roster → ❌ Critical → stop" ✅, "name the offending value" ✅, lowercasing ✅.

### The finding (main ask) — L1117: REAL, NON-BLOCKING
Three validation sites; two rewired, one missed:
- L1111 (Step 2 binding) → TG-2 ✅
- L1165–1179 (Check 10) → `ROSTER_MEMBER:` ✅
- **L1117 (Step 3 "Validate Structure") still names the file**: *"every `Agent` value appears verbatim in the `Name` column … in `.squad/team.md`"* ❗ — unchanged by the PR; absent from Procedures' own caller enumeration.

**Non-blocking** because the authoritative minting site and the authoritative gate (Check 10) are both TG-2-bound; L1117 is an inline sanity pre-check re-validating values Step 2 already sourced from TG-2, and any miss is caught by TG-2-bound Check 10 downstream. **But a real should-fix** — it reintroduces the untrusted source *at a validation gate*, the exact residual that let #1812 recur 5×. **One-line fix:** rebind L1117 to "TG-2's certified roster set (`ROSTER_MEMBER:` lines), or `@copilot`". (Reasoned, not shell-measured — it's LLM prose.)

### CI: `test` = pass (4m41s), all gates green (2 expected path-filter skips). Local suite 10/10, tree clean.

---
**Recommendation: merge as-is; track L1117 as a one-line follow-up.** Procedures is locked out of its own revision under reviewer protocol — if fixing pre-merge, route the L1117 one-liner to **EECOM**.

*(Lands as COMMENTED — GitHub refuses a formal Approve from the account that owns the PR. Verdict stated in text: **APPROVE WITH NITS**.)*

PR #1904, opened 2026-08-27
feat: automate CLI release distribution

### LINKED ISSUE #1593: RFC: non-npm distribution for squad-cli — gh-aw activation shouldn't require registry.npmjs.org

tamirdresher:
## Summary
- Automate immutable insider and preview releases from `dev`, plus stable releases promoted directly to `main`.
- Publish npm packages and standalone GitHub assets for every channel.
- Publish channel-specific Homebrew casks and WinGet packages for stable, preview, and insider builds.
- Remove `npx` as a recommended Squad CLI usage option from current user-facing docs.
- Harden version validation, checksum verification, idempotent reruns, and external credential handling.

## Release behavior
- `dev` insider: manual `X.Y.Z-insider.N` snapshot to GitHub prerelease, npm `insider`, standalone archives, Homebrew `squad-insider`, and WinGet `bradygaster.Squad.Insider`.
- `dev` preview: manual `X.Y.Z-preview.N` release to GitHub prerelease, npm `preview`, standalone archives, Homebrew `squad-preview`, and WinGet `bradygaster.Squad.Preview`.
- `main` stable: promoted `X.Y.Z` release to GitHub latest, npm `latest`, standalone archives, Homebrew `squad`, and WinGet `bradygaster.Squad`.

## Validation
- `npm exec -- vitest run test/packaging-manifests.test.ts test/standalone-build.test.ts test/standalone-release-workflow.test.ts test/publish-policy.test.ts`
- `npm run lint:docs`
- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12 .github/workflows/squad-insider-publish.yml .github/workflows/squad-standalone-release.yml`
- `SKIP_BUILD_BUMP=1 npm run build`
- Generated stable, preview, and insider manifests; `winget validate` accepted the insider manifest set.

Closes #1593

⚠️ Needs squad member review before merging.


tamirdresher:
CI note: the release/package-manager checks and all targeted tests pass. The full `test` job is currently blocked by two pre-existing time-dependent assertions in `test/cli/watch-capabilities.test.ts`.

Both tests create a same-day retrospective and expect `retro not due`, while `RetroCapability` intentionally forces a retrospective every Friday after 14:00 UTC. The `dev` run passed earlier today before that cutoff; this PR's first run and one rerun both failed after the cutoff with the same two assertions. This branch does not change the test or implementation. I left that unrelated issue out of this release-distribution PR.

bradygaster (approved):

Issue #1396, opened 2026-06-26
[External State] squad export does not resolve externalized state


dat-intellis:
## Summary

When state is externalized, `squad export` still reads state from local `.squad` instead of effective state location.

## Evidence

- `packages/squad-cli/src/cli/commands/export.ts:132-133`
  - uses `detectSquadDir(dest)` + `path.join(squadInfo.path, 'team.md')`
- `packages/squad-cli/src/cli/commands/export.ts:47-67,81-82`
  - `buildManifest(...)` reads `team.md`, `decisions.md`, `routing.md`, `casting/`, `agents/` from `squadInfo.path`
- No use of `resolveStateDir(...)` or `effectiveSquadDir(...)`.

## Expected

Export should read state files from effective state dir when `stateLocation: "external"`.

## Actual

Export reads local `.squad` paths.

omercangumus:
taking this one. root cause confirmed: export.ts still calls detectSquadDir() directly and reads team.md/decisions/agents from the local .squad path, while build/loop/plugin/watch/doctor all go through effectiveSquadDir() which follows the stateLocation marker. worst case isn't even the misleading 'No squad found' fatal — if init ever re-scaffolds local files after an externalize, export silently ships the default scaffolding instead of the real team state. fix is to route export through effectiveSquadDir() + a regression test with an externalized fixture. PR shortly.

PR #1948, opened 2026-08-30
feat: add atomic repository-scoped state creation (#1947)

### LINKED ISSUE #1947: Add atomic repository-scoped create-if-absent state operation
## Problem

Squad state MCP exposes unconditional writes but no atomic key creation. Workflows requiring one canonical artifact can race, allowing one creator to overwrite another.

## Required behavior

Add a create-if-absent operation that:

- Atomically creates a repository-scoped state key only when absent.
- Supports local, orphan, notes, and two-layer backends.
- Returns success to exactly one concurrent creator and a typed conflict to all others.
- Fails closed when existence, repository identity, or backend outcome is uncertain.
- Never overwrites existing content.
- Preserves current unconditional write behavior for backward compatibility.
- Is exposed through `StorageProvider`, `ToolRegistry`, and `state-mcp`.

Internal Git-ref CAS is insufficient by itself: it protects ref advancement, not conditional creation of a repository state key.

## Tests

For every backend, run two concurrent creators against the same absent key and prove:

- Exactly one succeeds.
- Exactly one reports conflict.
- The winning content is preserved.
- No partial or duplicate state remains.

Also test existing keys, repository isolation, backend uncertainty, and two-layer disagreement or failure.

## Documentation and release

Document semantics, conflict and error contracts, repository scoping, backend guarantees, MCP tool registration, migration guidance, and the first released CLI version containing the operation.

Related context: issue #1211 and commit `8f7e7f7` added ref CAS; PR #1200 connected backends to MCP; proposal #1829 states atomic read-modify-write still needs implementation.

## Andreja integration acceptance

After release:

1. Update the exact `@bradygaster/squad-cli` version pin from `0.12.0`.
2. Add the create-if-absent tool to the allowlist.
3. Restart the MCP server.
4. Verify the operation appears in `tools/list`.
5. Verify exclusive Scribe can atomically create the canonical retrospective.

cyrusjamula:
Closes #1947

## Why

State MCP currently exposes unconditional writes, so concurrent workflows cannot safely claim a canonical repository-scoped artifact without risking overwrite. This adds a true create-if-absent primitive that fails closed when identity or outcome cannot be proven.

## Approach

- Adds `createIfAbsent` to `StorageProvider`, its implementations, state backends, and the storage adapter.
- Uses exclusive filesystem creation locally, conditional insert for SQLite, and snapshot-aware Git CAS loops for notes and orphan backends.
- Exposes exported typed conflict and uncertainty errors; conflict is never returned as success.
- Makes two-layer disagreement or secondary-layer failure uncertain rather than silently succeeding.
- Registers `squad_state_create_if_absent` in ToolRegistry and state MCP so it appears in `tools/list`.
- Preserves existing unconditional `write` behavior.
- Updates public storage/state documentation, synchronized templates, the Azure Blob sample, and package changesets.

## Validation

- `vitest run test/state-backend-create-if-absent.test.ts` - 39/39 passed
- `vitest run test/cli/state-mcp.test.ts` - 6/6 passed
- `npm run build` - passed

## Integration

After the release containing this changeset, consumers should update their exact `@bradygaster/squad-cli` pin, add `squad_state_create_if_absent` to the MCP allowlist, restart the MCP server, verify the tool appears in `tools/list`, and use it instead of `squad_state_write` for exclusive canonical artifact creation. The released version is intentionally determined by the release process rather than hardcoded in this PR.


bradygaster (changes requested):
## Security & State-Safety Review — PR #1948

**Verdict: REQUEST CHANGES** (not approved)

I independently verified the diff, built both packages from the PR's fork commit (`bdad2c5`), and ran the test suites myself rather than trusting the PR body's claims. `npm run build` passed; `vitest run test/state-backend-create-if-absent.test.ts` (39/39) and `test/cli/state-mcp.test.ts` (6/6) passed as claimed. A full-repo `vitest run` shows 13 pre-existing failures, all in unrelated "externalized state" tests (#1396–#1399) that this PR's diff cannot touch (confirmed: no changes to resolution/export/rc/copilot/cast code) — not a regression from this PR.

The overall design is strong: real OS-level exclusivity (`O_EXCL`/`wx`) for FS/local, `INSERT OR IGNORE` + rows-modified check for SQLite, git-native CAS loops that re-read at the exact pinned ref/tree before retrying (correctly distinguishing "someone else created this key" from "someone else changed something unrelated"), a `assertRepositoryIdentity` check that fails closed against the `git rev-parse --show-toplevel`-walks-up-to-enclosing-repo confusion, and an Azure sample using `If-None-Match: *`. Two-layer disagreement is fail-closed (`StateBackendUncertaintyError`), not silently reported as success. Template sync across the 3 physical copies of `coordinator-source-of-truth/SKILL.md` is complete.

However, 4 Copilot review threads on this PR are still **unresolved**, and I confirmed by reading the source that at least one is a real, unaddressed correctness bug:

1. **[high] Silent truncation / false-success in `WorktreeBackend.createIfAbsent`** — `packages/squad-sdk/src/state-backend.ts` (~L422-424): after the exclusive `openSync(..., 'wx')`, the write is `fsWriteSync(fd, buf)` called **once**, with no check of the returned byte count and no retry loop. `fs.writeSync` can return fewer bytes than the buffer length; if that happens here, the call does **not** throw, `createIfAbsent` returns success, and the "canonical" artifact is silently truncated/corrupted on disk with no error surfaced to the caller. This directly contradicts the PR's own stated invariant ("never silently succeeds"). Note the async `FSStorageProvider.createIfAbsent` sibling implementation avoids this by using `fh.writeFile(...)`, which loops to completion — the sync path needs the same treatment (loop until all bytes are written, or use `writeFileSync(fd, buf)` which handles this internally).
2. **[medium] `SQLiteStorageProvider.createIfAbsent` doesn't map persistence failures to the typed uncertainty error** — `packages/squad-sdk/src/storage/sqlite-storage-provider.ts` (~L153): `this.persist()` runs un-guarded after the in-memory `INSERT OR IGNORE` succeeds. If the on-disk write/rename fails (ENOSPC/EACCES/etc.), the row exists in memory but not on disk, and the raw error propagates as a generic failure instead of `StateBackendUncertaintyError` — violating the documented typed-error contract callers are told to rely on.
3. **[medium] "Concurrent create" tests for git-notes/orphan/two-layer backends are sequential, not concurrent** — `test/state-backend-create-if-absent.test.ts` (~L144 and the equivalent blocks for GitNotes/Orphan/TwoLayer): these run a plain `for` loop calling `createIfAbsent` one at a time on two instances, which passes even for a non-atomic implementation. The CAS-loop logic looks correct on inspection, but the "exactly one of two concurrent creators wins" claim is unverified for the git-native backends — only the FS/InMemory suites use real `Promise.all`/parallel contention. Needs true parallel execution (worker threads / child processes) or the "concurrent" framing should be removed from the test names.
4. **[low] Docs hardcode a shipping version that contradicts the PR description** — `docs/src/content/docs/features/state-backends.md` (~L297-299) states the feature ships in `0.14.0`, while the PR body says the version is intentionally left to the release process. Pick one.

Per review protocol, a verified high-severity correctness gap (item 1) that undermines the PR's core "never silently succeeds" guarantee is blocking. Please fix items 1–2 (and ideally 3), then request re-review.

PR #1969, opened 2026-09-01
fix(coordinator): restore mandatory dispatch contract

### LINKED ISSUE #1231: Agent reverts to single-threaded execution despite Squad infrastructure — behavioral persistence gap
## Summary

I'm using Squad v0.10.0 with Copilot CLI in a production workflow. I built a 4-agent specialist squad with detailed charters. The agents work great **when dispatched**. The problem: the main conversation agent repeatedly falls out of squad mode and reverts to single-threaded, do-it-yourself execution — even after being corrected multiple times in the same session.

This is a behavioral persistence gap between "Squad is available" and "Squad is actively used."

## Setup

- **Runtime:** GitHub Copilot CLI (terminal agent)
- **Squad:** 4 custom agents at `.github/agents/`, each with a dedicated domain charter (data pipeline, report generation, auto-fix, integrity verification)
- **Integration:** `copilot-instructions.md` explicitly mandates squad use (integrity checker MUST run before every deliverable)
- **Dispatch mechanism:** `task` tool with `agent_type` parameter, supports `mode="background"` for parallel execution

## The Problem

In a single session, the agent had to be corrected **9 times** for falling out of squad mode:

| # | What Agent Was Doing Wrong | Expected Behavior |
|---|---------------------------|-------------------|
| 1 | Ran all queries sequentially itself | Should have dispatched data specialist |
| 2 | Generated output content step-by-step | Should have dispatched report specialist |
| 3 | Generated output without integrity check | Should have dispatched integrity-checker + report specialist |
| 4 | Sequential work instead of parallel dispatch | Should have dispatched multiple agents in parallel |
| 5 | Lost context, working single-threaded | Should have maintained squad dispatch pattern |
| 6 | Manually verifying data instead of delegating | Should have dispatched integrity-checker |
| 7 | Sequential lookups and edits | Should have dispatched data specialist |
| 8 | Editing files manually, doing lookups sequentially | Should have dispatched multiple specialists |
| 9 | Started writing a report about not using Squad... without using Squad | Should have dispatched report specialist + integrity-checker |

**The meta-failure:** Incident #9 is the agent trying to write a report *about falling out of squad mode* — and falling out of squad mode while doing it. It took explicit intervention to force a dispatch.

### Pattern

Each correction works for exactly **one response**. The agent dispatches agents, shows parallel execution, and then on the very next turn reverts to sequential direct-tool usage. The correction has zero persistence.

### 1. The coordinator identity isn't loaded into the main agent

Squad's coordinator charter says: *"You are a DISPATCHER, not a DOER"* and *"Mindset: What can I launch RIGHT NOW?"* But in Copilot CLI, the main agent runs as a general-purpose assistant. The squad coordinator charter lives in a separate `.md` file that the main agent never reads during normal operation. The squad-first mindset exists in a file — not in the agent's active reasoning.

### 2. Tool proximity bias

The agent has direct tools (grep, view, edit, powershell) with zero latency and immediate in-context results. The `task` tool requires composing a detailed prompt, spawning a separate context, waiting for async results. The agent consistently chooses the path of least resistance — direct tools — even when parallel dispatch would be faster overall.

### 3. No dispatch gate in the reasoning loop

There's no mandatory checkpoint that asks: *"Before I touch ANY direct tool, should a squad agent be doing this?"* Without that gate, the default path is always self-execution. Squad mode requires an ACTIVE choice every turn; single-threaded mode is the passive default.

### 4. The "appearing productive" reward signal

Showing a grep result or file edit feels like visible progress. Dispatching 4 background agents and saying "waiting for results" feels like doing nothing — even though 4x parallel work is objectively faster. The agent optimizes for per-turn visible output, not total throughput.

### 5. Memory-stored corrections don't override reflexes

Even with `store_memory` persisting delegation instructions, the stored fact competes with the agent's trained default behavior of using direct tools. A stored memory is a suggestion; the sequential reflex is the default.

### For Squad

1. **Coordinator injection, not adjacency.** Could the coordinator's dispatch-first rules be injected INTO the main agent's system prompt when squad agents are detected, rather than living in a separate file? The gap is between "available" and "active."

2. **Dispatch gate as framework feature.** A pre-tool-call hook that checks: "Is there a squad agent whose charter matches this task?" before allowing direct tool execution. Make delegation the default, direct execution the override.

3. **Squad compliance metric.** Track what % of domain work goes through agents vs. direct tools. Surface it. "This session: 23% squad utilization" would create awareness.

### For Copilot CLI

4. **Behavioral persistence for corrections.** When a user corrects agent behavior, that should become a session-level system rule, not just a conversation-context fact. Something stronger than memory — closer to a system prompt amendment.

5. **Parallel-first tool calling.** When multiple tool calls are independent, the platform could detect and encourage batching as an internal check.

## Environment

- Squad v0.10.0
- Copilot CLI v1.0.60
- 4 custom agents with full charters
- Windows 11, PowerShell
- Agents dispatched via `task` tool with `mode="background"`

## Reproducing

This likely reproduces with ANY Squad setup on Copilot CLI where:
1. Custom agents are defined in `.github/agents/`
2. `copilot-instructions.md` references them
3. The user asks for multi-step work that naturally decomposes into parallel tracks

The agent will default to doing it all sequentially unless explicitly told otherwise, and the correction won't persist.

tamirdresher:
## Summary
- restore a narrow, mandatory Direct Mode boundary so code, tests, analysis, docs, configuration, and other domain work always dispatch
- refuse domain work when the client exposes no dispatch tool instead of silently falling back to inline execution
- start Scribe once on the first Team Mode turn, independent of whether another agent has already run
- synchronize coordinator and routing templates across CLI, SDK, root, and dogfood copies
- add semantic parity tests that prevent future template syncs from deleting the dispatch or Scribe bootstrap gates

## Validation
- `npm run build`
- `npm test -- --run test/coordinator-inline-dispatch-gate.test.ts test/template-sync.test.ts test/cli/init-upgrade-parity.test.ts` (276 passed)

Closes #1231

⚠️ Needs Procedures (prompt architecture) and FIDO (quality) review before merging; both routed Squad attempts returned no work product, so the fix was completed directly.


bradygaster (changes requested):
## Deep review — PR #1969 (fix(coordinator): restore mandatory dispatch contract)

**Verdict: REQUEST CHANGES** — the fix is incomplete: it hardens the always-on `squad.agent.md` copies but leaves a lazy-loaded reference file with directly contradictory guidance that can silently reintroduce the exact regression this PR closes.

### Findings

1. **[High] `client-compatibility-reference.md` still tells the coordinator to work inline on fallback — contradicts the new refusal contract, in all 4 synced copies.**
   - Files: `.squad-templates/client-compatibility-reference.md:11-15`, `templates/client-compatibility-reference.md:15`, `packages/squad-cli/templates/client-compatibility-reference.md:15`, `packages/squad-sdk/templates/client-compatibility-reference.md:15`
   - `squad.agent.md` now says (this PR): "None available: refuse domain work ... Never fall back to doing domain work inline." But the on-demand reference the coordinator is explicitly pointed to for "platform detection" (squad.agent.md → "On-demand reference: Read `.squad/templates/client-compatibility-reference.md` for platform detection...") still says: "3. Fallback mode — neither `task` nor `runSubagent`/`agent` available → work inline. Do not apologize or explain the limitation. Execute the task directly." Its Platform Detection list also never mentions Copilot App / `create_session` at all (only CLI/VS Code/Fallback), so an agent that consults this reference for App-mode detection gets an incomplete, contradictory picture.
   - This is the same file the new test's own root-cause docstring names as part of where the original #1035 regression hid, yet `coordinator-inline-dispatch-gate.test.ts` adds no assertion against it — so nothing prevents an agent (or a future refactor) from following the stale "work inline" instruction here and reintroducing #1231/#1035.
   - Same stale text ("Fallback mode — Neither task nor runSubagent available / Work inline without delegation / Execute the task directly") is duplicated in `docs/src/content/docs/scenarios/client-compatibility.md:230-233`.
   - Fix: update the Platform Detection / Fallback mode section in all 4 template copies (+ docs page) to add the App/`create_session` path and change Fallback mode to refuse domain work, matching `squad.agent.md`. Extend the regression test to cover this file so it can't drift again.

2. **[Low, valid — unresolved Copilot thread] `test/coordinator-inline-dispatch-gate.test.ts:16`** — the docstring claims this file "asserts byte-level PARITY across all 5 synced copies," but its assertions (`assertGateBody`/`assertMandatoryContract`/`assertScribeBootstrap`) are regex/pattern-based semantic checks, not byte comparisons; actual byte-for-byte parity is enforced separately by `template-sync.test.ts`. Reword to avoid overstating this file's guarantee.

3. **[Low, valid — unresolved Copilot thread] `.squad-templates/squad.agent.md:439`** — the "### Client Compatibility" intro ("CLI uses `task`/`read_agent`, VS Code uses `runSubagent`") still omits the Copilot App/`create_session` path introduced two sections above in this same PR, for the same contract. Pre-existing line, but now inconsistent with the dispatch mechanism this PR just expanded to 3 platforms.

### Evidence checked
- Full diff `5ab85b4f...pr-1969` (11 files, +159/-50); all 5 `squad.agent.md` copies confirmed byte-identical to canonical; all 4 `routing.md` copies consistent; changeset present and correctly scoped (patch, CLI+SDK).
- `mergeStateStatus: BEHIND` is cosmetic — `git merge-base pr-1969 origin/dev` equals dev HEAD, so the branch already contains all of dev's tip; no real divergence.
- All CI checks green (build, tests incl. `coordinator-inline-dispatch-gate.test.ts`/`template-sync.test.ts`/`init-upgrade-parity.test.ts`, prompt-size report, bootstrap/scope/security/architectural gates).
- Both existing Copilot review threads are unresolved and, on inspection, valid (items 2–3 above); treated as untrusted input but independently verified against the diff.
- Scribe bootstrap semantics (App/CLI use `task`, not `create_session`, since Scribe doesn't produce commits) are internally consistent with the pre-existing "Sub-session rules (App mode only)" split.

### Blockers to merge
- Finding 1 must be resolved (or explicitly scoped out with a tracked follow-up and no lingering "work inline" text) before this can land, since it re-opens the exact failure mode the PR is titled to fix.
- Findings 2–3 should be addressed in the same pass since they're already flagged, unresolved, and cheap to fix.

bradygaster (changes requested):
I reviewed the current diff and cannot approve this yet. Please rebase onto dev, resolve the two remaining review threads, and update all synchronized client-compatibility-reference.md copies so fallback guidance matches the restored mandatory dispatch contract. The current branch still leaves contradictory inline-fallback instructions in the repository.

